PR #1, opened 2015-07-27
Fix '=>' precedence.

MichaelHatherly:


MichaelHatherly:
Side question: I'm trying to get Atom with Julia going with these packages, presumably I need ``CodeTools``, ``Atom``, ``atom-julia-client``, ``language-julia``, and ``atom-ink``. Are there any others I've missed?

one-more-minute:
Yup, that's it. Don't hesitate to email me if you have questions and/or want to go have a go at hacking on this stuff, it'd be great to have your input.

MichaelHatherly:
Thanks, will do.

PR #2, opened 2015-07-27
Sort modules.

MichaelHatherly:
Very minor change, but displaying the module dropdown in alphabetical order fits better with other dropdowns such as the language picker. Thoughts?


one-more-minute:
Make sense, at the moment it seems like the order is pretty much random every time so this is a lot better.

PR #5, opened 2015-09-05
adds a mutex lock around the socket println. fixes #3

### LINKED ISSUE #3: socket `println` can get interleaved
I think this is one of the root causes of junolab/atom-julia-client#19. When I added a mutex to comm.jl as in the below code I'm seeing all the incoming messages properly separated by newlines.

It's pretty ugly and really should be an outgoing message queue or something, but I think it demonstrates the issue.

```
type Mutex
    locked::Bool
    cond::Condition

    Mutex() = new(false, Condition())
end

function lock(m::Mutex)
  while m.locked
      wait(m.cond)
  end
  m.locked = true
end

function unlock(m::Mutex)
    m.locked = false
    notify(m.cond)
end

global msgMutex = Mutex()
function msg(t, data)
  isactive(sock) || return
  lock(msgMutex)
  println(sock, json(c(t, data)))
  unlock(msgMutex)
end
```

ssfrr:
This shouldn't strictly be necessary, but is a workaround for JuliaLang/julia#10615

### DIFF
--- a/src/comm.jl
+++ b/src/comm.jl
@@ -1,4 +1,5 @@
 global sock = nothing
+global msgMutex = ReentrantLock()
 
 isactive(sock::Nothing) = false
 isactive(sock) = isopen(sock)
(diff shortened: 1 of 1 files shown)


one-more-minute:
Awesome, thanks for doing the digging on this issue. Hopefully JuliaLang/julia#12978 will make the workaround unnecessary.

PR #6, opened 2015-09-14
Use correct paths for include etc.

one-more-minute:
This change essentially ensures that the output of `@__FILE__` is correct. As an important side effect, the `include` function now works as it does when loading the file the first time – i.e. it loads files relative to the current file's directory.


spencerlyon2:
LGTM

PR #7, opened 2015-09-15
fix regexp in `splitlink` on Windows

Varanas:
should fix https://github.com/JunoLab/atom-julia-client/issues/30, but I don't know if I break something with that naive regexp. Though it seems to work with linux-style paths as well as Windows style paths.


Varanas:
Alright, so this regexp is just wrong :)
`r"^([^:]*:[^:]*)(?=:(\d*))?"` works if there is a second `:` in the path (so we could call fullpath and make it windows only) -- I'll figure something out tomorrow.

Varanas:
fixed with 2fbac8d.

PR #8, opened 2015-09-17
RFC: First stab at setting module for console eval

spencerlyon2:
This is the Julia-side of having the console evaluate in the module of the last open Julia buffer.

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -108,9 +108,10 @@ handle("evalrepl") do data
       elseif mode == "help"
         code = "@doc $code"
       end
+      mod = getthing(data["module"], Main)
       try
         withpath(nothing) do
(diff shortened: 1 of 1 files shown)


spencerlyon2:
Related to JunoLab/atom-julia-client#34

spencerlyon2:
This appears to work. Here's a screenshot taken by following these steps:

where before the first call to `current_module()` I was in `~/.julia/v0.4/Atom/src/Atom.jl`, then

spencerlyon2:
I think regardless of what we decide over on JunoLab/atom-julia-client#34, this is probably the right approach on the Julia side

Varanas:
This has been implementd, closing.

PR #11, opened 2015-10-30
better rendering of large Dicts

Varanas:
and a minor comsetic fix for large vectors.
This works towards fixing https://github.com/JunoLab/atom-julia-client/issues/54, but I would much prefer a more general solution to this problem.


Varanas:
Any objections? Otherwise I'll merge this tomorrow.

one-more-minute:
Looks good to me! In future we definitely could do with the lazy loading stuff from LT, but it's tricky to implement unfortunately.

PR #17, opened 2015-12-26
Add handlers and renderers for docs and methods

Varanas:
This is the Julia side for supporting docs-at-a-keystroke in Atom.

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -99,3 +99,13 @@ handle("eval-repl") do data
     end
   end
 end
+
+handle("docs") do data
+  result = @errs include_string("@doc $(data["code"])")
+  @d(:result => render(Editor(), result))
+end
+
+handle("methods") do data
+  result = @errs include_string("methods($(data["code"]))")
+  @d(:result => render(Editor(), result))
+end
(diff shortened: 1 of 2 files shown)


src/eval.jl
```diff
@@ -99,3 +99,21 @@ handle("eval-repl") do data
     end
   end
 end
+
+handle("docs") do data
+  @dynamic let Media.input = Editor()
+    result = withpath(getpath(data)) do
```
spencerlyon2: Why do we need the `withpath(getpath(` here?

Varanas: Because I have no idea how the docsystem works and just copied that from above.
It's probably unnecessary, but I'm not sure.

one-more-minute:
Ok, so you're just treating doc requests as if the user had evaluated `@doc X`? Sounds like a reasonable approach as a first pass.

src/eval.jl
one-more-minute: No, I don't think it should be necessary in this case, nor the Media.input binding. Also, on Julia 0.4 you can switch `@d` for just `d`.

Varanas:
Yeah. I don't have any specific ideas for another approach, actually, besides the possibility to open a new Blink or browser window to display the online docs (like matlab does).

PR #18, opened 2015-12-28
Handle undef entries in vectors

Varanas:
Fix for https://github.com/JunoLab/atom-julia-client/issues/44.

Since we aren't doing any special rendering with arrays, they work anyway. It would be really cool to have this work for arbitrary arrays, but I couldn't figure out how to do that without constructing an array of the same size as the array to be printed right now.
So when we implement nice Array printing, this should be overhauled.

Also: Right now undef elements are printed as `"#undef"`, but that's only a small cosmetic problem, methinks.

### DIFF
--- a/src/display/objects.jl
+++ b/src/display/objects.jl
@@ -59,3 +59,19 @@ end
 end
 
 @render i::Inline x::Number Text(sprint(show, x))
+
+function handleundefs(X::Array)
+  Xout = Array{Union{ASCIIString, eltype(X)}}(size(X))
+  for i = 1:prod(size(X))
+    Xout[i] = isdefined(X, i) ? X[i] : "#undef"
+  end
+  Xout
+end
+
+function handleundefs(X::Array, inds)
+  Xout = Array{Union{ASCIIString, eltype(X)}}(size(X))
+  for i in lininds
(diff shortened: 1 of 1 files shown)


spencerlyon2:
What if we had some higher-dimension Array?

Couldn't this same approach work for `Array{T,N}` for any `N`?

Varanas:
Then this doesn't work ;)
Right now, arrays are printed as they are in the REPL, so none of the issues we have with vectors exist. Of course, an expansion of the inline result stuff is in order, and then we'll have to deal with the same probelm for arrays.

Expanding this approach to arrays works, of course, but isn't trivial if you want to only create small arrays, imo. If someone has any great ideas, then I'm all ears ;)

spencerlyon2:
Oh I didn't realize that we are currently handling `N>1` the same as the REPL. 

I think that this fix looks good then, although it makes me wonder why we didn't just use REPL output for `Vector` also?

Varanas:
Well, it does look better if we style it ourselves:
unbenannt

The plan is to move ahead with something like suggested in #12, so we probably will implement all of that stuff in Atom anyways.

spencerlyon2:
Sounds good. Let's merge and review for `N>1` if/when we need to

Varanas:
Alright, thanks for having a look! :)

PR #19, opened 2015-12-28
Correctly render errors with multiline header

Varanas:
by splitting them on `\n` and only displaying the first line in the header. 

Example: 
```julia
Vector(AbstractString,4)
#ERROR: MethodError: `convert` has no method matching convert(::Type{Array{T,1}}, ::Type{AbstractString}, ::Int64)
#This may have arisen from a call to the constructor Array{T,1}(...),
#since type constructors fall back to convert methods.
````

### DIFF
--- a/src/display/errors.jl
+++ b/src/display/errors.jl
@@ -3,7 +3,7 @@ type EvalError
   bt
 end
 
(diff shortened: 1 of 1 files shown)


one-more-minute:
big ol' retrospective +1

PR #21, opened 2016-01-10
fix for not properly dealing with modules 

Varanas:
The `eval(parse(mod))` thing is obviously a bit scary, but I dunno how else to construct a `Module` from a string... Any ideas anyone? @one-more-minute 

Also, this is on top of #20.


Varanas:
Now incorporated in #20.

PR #24, opened 2016-01-25
add MacroTools and Requires explicitly to REQUIRE

tkelman:
if they're being directly used, they should be directly depended on,
rather than assuming they will be present as transitive deps of other packages


Varanas:
Not quite sure why this sat here unmerged for as long as it did, but thanks!

PR #26, opened 2016-02-22
add clear-workspace msg handler

jkroso:
to enable https://github.com/JunoLab/atom-julia-client/pull/115

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -17,3 +17,8 @@ end
 handle("exit") do
   exit()
 end
+
+handle("clear-workspace") do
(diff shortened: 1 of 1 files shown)


MikeInnes:
Looks good, but I'm not sure about the `clear_workspace` thing. How about just calling `workspace()` directly, and you can always redefine how it works by redefining `Atom.handle("clear-workspace") do ...` in `.juliarc.jl`?

jkroso:
Oh good point!

MikeInnes:
Looking good, thanks again!

PR #27, opened 2016-03-03
enable rendering of compose contexts

pfitzseb:
ref: https://github.com/JunoLab/Media.jl/pull/6

compose

This is very much preliminary support only.

I'm not sure about adding the `render(::Editor, plot::Compose.Context)` method, but Atom slows down way too much when displaying the entire result  inline. Also, this is similar to how Gadfly plots behave. 
Dunno what the best way forward is for these cases where we have side effects to evaluating something...

### DIFF
--- a/src/display/plots.jl
+++ b/src/display/plots.jl
@@ -28,3 +28,13 @@ end
   @render Inline v::Vega.VegaVisualization Text("Vega Visualisation")
   @render Clipboard v::Vega.VegaVisualization Text("Vega Visualisation")
 end
+
+@require Compose begin
+  @render PlotPane img::Compose.Context begin
+    HTML(stringmime(MIME"image/svg+xml"(), img))
+  end
+
+  @render e::Editor img::Compose.Context begin
+    Text("Compose.Context(...)")
+  end
+end


MikeInnes:
Looks good – does it work OK with a dark theme too?

Can you clarify what you mean by

> Dunno what the best way forward is for these cases where we have side effects to evaluating something...

?

pfitzseb:
Well, the background is transparent... Not sure if we should set a white background by default.

The side effects thing was about how to display something like `Compose.Context`s: Plotting is a side effect the user may actually not want without explicitly requesting it, e.g. calling `display` on it or clicking on the result (which could be a static preview or something).

MikeInnes:
My thinking is that it just means that Compose contexts are always displayed in a consistent way. Someone working in the repl might not think to explicitly add a white background, and then their work is going to look bad in most people's Atom setup. Perhaps there are reasons not to do that, though.

I see what you mean about side effects. See also https://github.com/JuliaLang/atom-language-julia/issues/61. In general we've set things up to lean towards the current model, and I don't know how else we could do things, so we should probably just stick with it for patches like this. But I'm also very open to proposals for how we can do things differently.

PR #29, opened 2016-03-17
WIP: Tests

pfitzseb:


pfitzseb:
@MikeInnes Could you enable travis for Atom? I don't have the necessary admin rights for that ;)

MikeInnes:
I wouldn't say no to this, though as a rule I suspect that the integration tests in julia-client are more worthwhile to spend time on – since they test for correct behaviour independent of implementation. Testing the implementation can amount to specifying it twice, which isn't necessarily bad but (IMO) doesn't carry as much benefit. (I think we might have had this conversation before though.)

Anyway, will go ahead and enable travis!

tkelman:
bump. julia side unit tests are going to be necessary for proper cross version testing.

tkelman:
any recent progress?

pfitzseb:
Superseded by #60.

PR #41, opened 2016-06-05
better shell exec in console

pfitzseb:
fixes https://github.com/JunoLab/atom-julia-client/issues/177

would be great if someone could test this on something not-windows, but it *should*  work everywhere ;)


MikeInnes:
Ah you went ahead and implemented this. Great, this is clearly going to be an improvement and if there're issues we can fix them as they come up.

PR #42, opened 2016-06-08
inline markdown rendering

pfitzseb:
~~Short~~ design question: How do we want to handle different "rendering styles"?

Problem:
When rendering documentation in either a `Result > Tree > Children` or in a `InlineDoc` element, I don't want to create a `Tree` or `SubTree` for multiline Markdown but rather render the whole thing. But when rendering it inline, i.e. as a toplevel result, I want the implementation proposed here and create a `Tree`.

We should probably just create a new type and specialize our rendering methods on that to get that behaviour -- first I was thinking of something like `Underline`, but maybe something like `Verbose` fits better? 
If we do this, how would it integrate with the rest of the display system on the Julia side?

### DIFF
--- a/src/display/objects.jl
+++ b/src/display/objects.jl
@@ -80,3 +80,9 @@ end
 @render i::Inline xs::Tuple begin
   span(c("(", (@_ xs map(x->render(i, x), _) interpose(_, ", "))..., ")"))
 end
+
+@render i::Inline md::Base.Markdown.MD begin
+  mds = CodeTools.flatten(md)
+  length(mds) == 1 ? Text(chomp(sprint(show, MIME"text/markdown"(), md))) :
+                     Tree(Text("MD"), [HTML(sprint(show, MIME"text/html"(), md))])
+end


MikeInnes:
Having a simple `type Underline end` and overloading display methods for that when rendering docs seems like a reasonable approach. What we also do in the console is just to expand the top-level by default, which I imagine would work for inline results as well.

I'm not sure why you'd want to avoid the tree elsewhere though. If you have a `Vector{MD}`, you want the elements to be both compact and clearly distinguished, right?

pfitzseb:
md

In the above picture I'd rather not have `Tree("MD", [content])` but rather only `content` rendered as markdown. So it's just a few cases where we wouldn't want the compact form I guess.

MikeInnes:
Not sure I agree on that. Myself, usually when I click on a function its to get to the methods, so I like having things compact and getting the choice by default.

Either way, I don't think anything that special is needed for that case; the function display tree explicitly includes the markdown docs, so it could explicitly choose how it's rendered as well; whether that's with a `render_verbose` method or whatever.

pfitzseb:
Alright, fair enough. Do you have a good idea as to the text in the tree's `head`? 
I don't really like having `MD` there, but we can't very well just render the first line of the Markdown since that'd mean having arbitrary heights.

PR #46, opened 2016-07-06
latest thing evaled in the console assigned to `ans`

pfitzseb:
not sure if that change to the evaluation logic is correct, but I think it should be; also not sure if this is really desirable...

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -103,7 +103,7 @@ handle("evalrepl") do data
       @run begin
         @dynamic let Media.input = Console()
           withpath(nothing) do
-            render(@errs eval(mod, :(include_string($code))))
+            render(@errs include_string(mod, "ans=$code"))
           end
         end
       end
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -103,7 +103,7 @@ handle("evalrepl") do data
       @run begin
         @dynamic let Media.input = Console()
           withpath(nothing) do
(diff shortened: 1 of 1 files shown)


MikeInnes:
The change to eval logic looks right, but it might be best to avoid changing the code string itself. i.e. we could have `eval(mod, :(ans = include_string(...`. Not sure if there are any specific problems with `"ans=$..."` but it seems safer.

pfitzseb:
Hm, I think I tried that but the scoping wasn't quite right. Not sure though (it seems like it should work if I think about it now) and I don't really have time to check now (in China at the moment ;)), so feel free to make that change and close this PR.

MikeInnes:
I'll take a look if I get some time. Enjoy China!

MikeInnes:
One small thought about how this: it might be confusing to have different `ans`s in different modules, but we could do something like

```julia
result = ... do the eval ...
isdefined(mod, :ans) && eval(mod, :(import Main.ans))
eval(mod, :(ans = $result))
```

Probably a little fiddlier than that but you get the idea. Not urgent anyway.

MikeInnes:
I was worried about the edge cases but it turns out the repl doesn't handle them either:

```julia
julia> const ans = 1
1

julia> 2
WARNING: redefining constant ans
2

julia> :foo
ERROR: invalid redefinition of constant ans
```

PR #49, opened 2016-08-12
RFC: scrub backtrace before showing

KristofferC:
This is just an idea for scrubbing the backtraces before showing it to the user. Right now, the unnecessary junk we want to get rid of is

```jl
 in include_string(::String, ::String) at ./loading.jl:380
 in include_string(::String, ::String, ::Int64) at /home/kristoffer/.julia/v0.5/CodeTools/src/eval.jl:28
 in include_string(::Module, ::String, ::String, ::Int64, ::Vararg{Int64,N}) at /home/kristoffer/.julia/v0.5/CodeTools/src/eval.jl:32
 in (::Atom.##89#92{String,Int64,String})() at /home/kristoffer/.julia/v0.5/Atom/src/eval.jl:39
 in withpath(::Atom.##89#92{String,Int64,String}, ::String) at /home/kristoffer/.julia/v0.5/Requires/src/require.jl:37
 in withpath(::Function, ::String) at /home/kristoffer/.julia/v0.5/Atom/src/eval.jl:45
 in macro expansion at /home/kristoffer/.julia/v0.5/Atom/src/eval.jl:56 [inlined]
 in (::Atom.##88#91{String,Int64,String})() at ./task.jl:54
```

So what I am doing now is just finding the first `include_string` and remove everything after that. This is probably not strictly correct since a user could maybe have a non junk `include_string` call in his/her code.

I am not sure exactly what is best here but I am rolling with this locally and love that the backtraces are not crazy big anymore...

Maybe there could be some heuristic for example finding the first `include_string` in `CodeTools` and then remove all stack frames from one above that to the end.

### DIFF
--- a/src/display/errors.jl
+++ b/src/display/errors.jl
@@ -16,9 +16,12 @@ function splitlink(path)
 end
 
 # TODO: don't work on the text
-# TODO: clip traces so Atom/CodeTools doesn't show up
 
 function btlines(bt, top_function::Symbol = :eval_user_input, set = 1:typemax(Int))
+  include_ind = Base.REPL.findfirst(addr->Base.REPL.ip_matches_func(addr, :include_string), bt)
+  if include_ind != 0
+    bt = bt[1:include_ind-1]
+  end
   @_ begin
     sprint(Base.show_backtrace, bt)
     split(_, "\n")


KristofferC:
Also, the functions I'm using are quite new non exported Julia functions so won't work on 0.4 as is.

pfitzseb:
What do you think of something like
```julia
function btlines(bt, top_function::Symbol = :eval_user_input, set = 1:typemax(Int))
  trace = Base.stacktrace(bt)
  t_trace= map(trace) do frame
    (sprint(Base.StackTraces.show_spec_linfo, frame),
    string(frame.file),
    frame.line,
    frame.inlined ? " [inlined]" : "")
  end
  internal_ind = findfirst(t_frame -> ismatch(r"CodeTools", t_frame[2]), t_trace)
  t_trace[1:internal_ind-2]
end
```
which also addresses the other TODO.

KristofferC:
Definitely better! I guess just matching `"CodeTools"` could potentially match something spurious but that seems quite unlikely.

pfitzseb:
True. Slightly better heuristic:
```julia
  internal_ind = findfirst(t_trace) do t_frame
    ismatch(r"CodeTools(\\|/)src(\\|/)eval\.jl", t_frame[2]) && 
    ismatch(r"include_string", t_frame[1])
  end
```

MikeInnes:
Big :+1:!

To make this reliable (even when working on CodeTools.jl itself in Atom) you probably want to find the *last* include string rather than the first and cut from there. I think that finding the last include_string at loading.jl would be the most reliable option.

KristofferC:
Yeah, `findlast` is definitely correct now when the hueristic is better. The reason I had to use `findfirst` in this PR was because I just crudely matched `:include_string`. @pfitzseb version with `findlast` should work perfectly.

pfitzseb:
So:
```julia
  internal_ind = findlast(t_trace) do t_frame
    ismatch(r"loading\.jl", t_frame[2]) && 
    ismatch(r"include_string", t_frame[1])
  end
```
Looks good.

MikeInnes:
Makes total sense – looking forward to having this in.

As an aside, this code really needs refactoring, especially with the new Stacktraces module in base. But that's another issue.

KristofferC:
Sounds good. Will you handle it @pfitzseb?

pfitzseb:
Sure, can do.

@MikeInnes Do we want to switch to the [version directly operating on the stackframes](https://github.com/JunoLab/Atom.jl/pull/49#issuecomment-239396177) or keep the old regex approach for now?

KristofferC:
One thing I noticed now is that you still get the the "expand arrow" on the error even if it is just one line long:

image

You feel a bit tricked when you try to expand it and it is nothing there :P

MikeInnes:
Sure, if you want to some refactoring on this file I'm definitely not going to say no.

MikeInnes:
@KristofferC Well I guess we haven't noticed that before as we haven't had nicer stacktraces :) Noted and I'll get on this soon.

PR #53, opened 2016-08-17
remove undefined options to fix @which

dlfivefifty:
Fixes https://github.com/JunoLab/atom-julia-client/issues/214#issuecomment-240312544


pfitzseb:
Thanks!

PR #55, opened 2016-08-23
RFC: hide function argument types in backtrace

KristofferC:
Always scrolling to the right to find the line info in the error box gets a bit tedious. For interactive use like in Atom I believe that showing all the types that go into the functions in the backtrace is overkill. It is simple to click the link and look at the actual function if one wants to.

### DIFF
--- a/src/display/errors.jl
+++ b/src/display/errors.jl
@@ -19,7 +19,7 @@ function btlines(bt)
   isempty(bt) && return []
   trace = Base.stacktrace(bt)
   t_trace = map(trace) do frame
-    (sprint(Base.StackTraces.show_spec_linfo, frame),
+    (split(sprint(Base.StackTraces.show_spec_linfo, frame), '(')[1],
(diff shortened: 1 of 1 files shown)


MikeInnes:
Interesting idea, I'm not against at least trying this out for a while to see how it feels. Can you post a screenshot?

KristofferC:
image

image

Just an idea that I personally like but maybe people find the argument types valuable.

Noticed now that identical stack frames should probably be compressed into one entry and a `repeats n times` like the REPL does.

MikeInnes:
Ok, let's give this a go.

Definitely agree about the repetition; we also need to figure out a reasonable way to handle stack overflows, as opposed to crashing the editor with MiB of stack trace.

KristofferC:
I am reversing a bit on this... Sometimes it is convenient to not see the arguments but if people use it to report bugs it can be annoying for the julia devs.

PR #57, opened 2016-09-12
move progress and view to Juno.jl

pfitzseb:


MikeInnes:
Looks like this isn't relevant after your more recent progress changes, but let me know if that's not the case.

pfitzseb:
Right, just forgot to close this PR.

PR #65, opened 2016-10-06
Respect semicolon

TsurHerman:
Editor and Console respect semi-colon at the end. 
errors are still being rendered if they occurred while evaluating

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -20,6 +20,17 @@ function getmodule(data, pos)
   getthing("$main.$sub", getthing(main, Main))
 end
 
+function ends_with_semicolon(line)
+    match = rsearch(line, ';')
+    if match != 0
+        for c in line[(match+1):end]
+            isspace(c) || return c == '#'
+        end
+        return true
+    end
+    return false
+end
+
 handle("module") do data
   main, sub = modulenames(data, cursor(data))
   return d(:main => main,
@@ -34,9 +45,10 @@ end
 
 isselection(data) = data["start"] ≠ data["stop"]
 
-macro errs(ex)
+macro errs(ex , suppress = false)
   :(try
-      $(esc(ex))
+      ans = $(esc(ex))
+      $(esc(suppress)) ? nothing : ans
     catch e
       EvalError(isa(e, LoadError) ? e.error : e, catch_backtrace())
     end)
@@ -54,7 +66,7 @@ handle("eval") do data
   lock(evallock)
   result = @dynamic let Media.input = Editor()
     withpath(path) do
-      @errs include_string(mod, text, path, line)
+      @errs include_string(mod, text, path, line) ends_with_semicolon(text)
     end
   end
   unlock(evallock)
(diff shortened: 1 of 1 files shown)


src/eval.jl
```diff
@@ -20,6 +20,17 @@ function getmodule(data, pos)
...
+function ends_with_semicolon(line)
+    match = rsearch(line, ';')
+    if match != 0
+        for c in line[(match+1):end]
+            isspace(c) || return c == '#'
+        end
+        return true
+    end
+    return false
+end
```
MikeInnes: Is this the same function that exists in Base? If so, it'd be better to just call that.

src/eval.jl
```diff
@@ -54,7 +66,7 @@ handle("eval") do data
   lock(evallock)
   result = @dynamic let Media.input = Editor()
     withpath(path) do
-      @errs include_string(mod, text, path, line)
+      @errs include_string(mod, text, path, line) ends_with_semicolon(text)
```
MikeInnes: This feels unclean to me – `@errs` should only have one clear responsibility. Instead we could check whether the result should be suppressed just before rendering it, for example.

src/eval.jl
pfitzseb: It is, I think. But needs a bit of tweaking anyways -- the base method doesn't handle linebreaks and stuff while we need to.

MikeInnes: Really the repl should handle linebreaks and stuff too, given that it supports multiline input.

We may not be able to accept this for the editor unless it works reliably in the vast majority of cases, since input is more varied there – matching the repl in the console is fine though.

src/eval.jl
TsurHerman: This is what I did in the beginning but then you would want to signal the renderer not only if there 
was a semi-colon but also if there was an error.

And then @errs needs to return not only the result(error message or value) but also whether it was an error. maybe the error message have a special type that can be checked .. then it can be added as a condition for rendering , this will also expose the whole logic in code:

so is there a special type for the error message?

regarding evaluating multi-line in the editor, it works just as before but suppresses the output if the last line ends with semi-colon

TsurHerman:
How is it now?  .. should I submit a new PR ? I am new to this things..

src/eval.jl
MikeInnes: > regarding evaluating multi-line in the editor, it works just as before but suppresses the output if the last line ends with semi-colon

That's not quite true actually. For example:

```julia
Base.REPL.ends_with_semicolon("""
  function foo()
    foo; # bar
  end
  """) # -> true
```

So the output of that code block would be suppressed, which could be confusing.

However, I think this is sufficiently unlikely to cause problems in practice that we can just wait and fix it if it does.

src/eval.jl
```diff
@@ -20,6 +20,8 @@ function getmodule(data, pos)
   getthing("$main.$sub", getthing(main, Main))
 end
 
+
+
```
MikeInnes: Some extra whitespace has crept in here, please remove it

src/eval.jl
pfitzseb: It does -- for some reason I though that `isspace('\n') == false`.

src/eval.jl
pfitzseb: `Base.REPL.ends_with_semicolon(filter(x->!isempty(x), split(text, '\n')[end]))`?

MikeInnes: Yes, @TsurHerman let's apply ends_with_semicolon to the last line only. Note we can use `split(..., keep=false)` as opposed to the filter.

pfitzseb:
LGTM.

MikeInnes:
Great. Thanks @TsurHerman for the hard work and @pfitzseb for the review.

TsurHerman:
@MikeInnes 
Hard work :-) .. I do this to get my mind off my hard work

PR #66, opened 2016-10-20
Profiler

MikeInnes:
Basic profiler setup. Also some improvements to error rendering.

I haven't done any of the flashy gui integration yet (flame charts etc) but Juno's native tree view actually makes for a pretty decent viewer without any other modifications:

screenshot 2016-10-20 20 08 09

### DIFF
--- a/src/Atom.jl
+++ b/src/Atom.jl
@@ -16,6 +16,7 @@ include("misc.jl")
 include("frontend.jl")
 include("utils.jl")
 include("debugger/load.jl")
+include("profiler/profiler.jl")
 
 include("blink/BlinkDisplay.jl")
 @reexport using .BlinkDisplay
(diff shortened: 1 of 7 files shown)


pfitzseb:
Looks good!
Two small things:
 - This errors with `ArgumentError("reducing over an empty collection is not allowed")` when no samples have been collected, but should give a messsage like `Profile.print()` does.
 - `WARNING: symbol is deprecated, use Symbol instead.`

MikeInnes:
Ok, I think I've addressed those points.

pfitzseb:
I've pushed two small fixes, one for Windows support and the other for syntax highlighting in the profile tree (should probably create some methods or string macros for that, i.e. `Atom.Syntax.func(x) = Hiccup.span(".syntax--support.syntax--function", x)` or similar).

Looks good to merge, imho. And big 👍 for the new error printing code... :)

MikeInnes:
Great changes!

PR #76, opened 2016-12-26
render evalall result into console

pfitzseb:
This will show the result of the last block in the console, consitent with `include()`ing files in the REPL.

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -78,7 +78,10 @@ handle("evalall") do data
   @dynamic let Media.input = Editor()
     withpath(path) do
       try
-        include_string(mod, code, path)
+        result = include_string(mod, code, path)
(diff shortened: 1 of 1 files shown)


MikeInnes:
I'm not sold on this change. "run all" is meant to mean "run this script", and running scripts doesn't include any notion of a return value since there's nowhere to return to anyway.

`include` happens to return a value, but that's specific to `include`, which isn't involved at a user level when this command is run. People who get confused by this will also get confused by `julia script.jl` so there's not much we can do about that. Meanwhile, people who want this will eventually want to display every block, or the second-last block, or something else; we can solve that more generally by telling people to use normal show methods in the script, and it will then work as expected everywhere.

pfitzseb:
Agreed. (Wasn't very convinced myself anyways, which is why this is a PR in the first place).

PR #79, opened 2017-03-13
Add a single function to preprocess code

bramtayl:
for easy user customization

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -1,7 +1,7 @@
 using CodeTools, LNR, Media
 import CodeTools: getthing
 
-preprocess(x) = x
+preprocess = identity
 ends_with_semicolon(x) = Base.REPL.ends_with_semicolon(split(x,'\n',keep = false)[end])
 
 LNR.cursor(data::Associative) = cursor(data["row"], data["column"])
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -1,6 +1,7 @@
 using CodeTools, LNR, Media
 import CodeTools: getthing
 
+preprocess(x) = x
 ends_with_semicolon(x) = Base.REPL.ends_with_semicolon(split(x,'\n',keep = false)[end])
 
 LNR.cursor(data::Associative) = cursor(data["row"], data["column"])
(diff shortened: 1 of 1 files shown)


src/eval.jl
```diff
@@ -1,6 +1,7 @@
 using CodeTools, LNR, Media
 import CodeTools: getthing
 
+preprocess(x) = x
```
ChrisRackauckas: Should this be `preprocess = identity` so that way it's easier to reference?

bramtayl: I mean that sounds find to me but I'm not sure if it makes a difference? 

I was planning on using this personally to replace this with
```
preprocess(text) = *("import ChainRecursive\nChainRecursive.@chain begin\n", text, "\nend")
```
but
```
myprocess(text) = *("import ChainRecursive\nChainRecursive.@chain begin\n", text, "\nend")
preprocess = myprocess
```
would work just as well

ChrisRackauckas: Yes, that would still work. It would have the added benefit though that you could check if the preprocessor has been changed by checking `Atom.preprocess == identity`. If it's a unique function `preprocess(x) = x`, you don't have a reference to check against.

bramtayl: I tried running
```
import Atom
Atom.preprocess(text) = string("import ChainRecursive\nChainRecursive.@chain begin\n", text, "\nend")
```
from within the Juno but it didn't change behavior immediately. Will this change post 265?

ChrisRackauckas: Something might be weird due to 265 and precompilation.

bramtayl: Is there a way to retrigger precompilation?

sglyon: `Base.compilecache`?

MikeInnes:
I think you can basically do this already, and in a more general way. e.g.

```julia
process = Atom.handlers["eval"]
Atom.handlers["eval"] = function (data)
  data[:code] = ...
  process(data)
end
```

bramtayl:
Oh yeah that looks great. It might be good if `Atom.handlers["eval"]` held a named function so that the default can't get lost.

Issue #91, opened 2017-06-24
Remove Gallium/ASTInterpreter Dependencies?


ssfrr:
ASTInterpreter is forcing AbstractTrees to downgrade to a version that's not compatible with 0.6. See Keno/ASTInterpreter.jl#69 for deets.

pfitzseb:
Yeah, this definitely makes sense.

PR #92, opened 2017-06-25
remove dependency on gallium and astinterpreter

pfitzseb:
This actually works without any changes in code, since Gallium and ASTInterpreter always were loaded dynamically; the code that does this checks the Julia version, so we can't get a "missing package" warning from this.

Fixes https://github.com/JunoLab/Atom.jl/issues/91.


ssfrr:
:+1: Thanks!

PR #93, opened 2017-07-01
allow rendering of console results in different displays

pfitzseb:
amazing that no one has complained about this yet...

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -52,7 +52,7 @@ handle("eval") do data
     unlock(evallock)
 
     display = Media.getdisplay(typeof(result), Media.pool(Editor()), default = Editor())
-    !isa(result,EvalError) && ends_with_semicolon(text) && (result = nothing)
+    !isa(result, EvalError) && ends_with_semicolon(text) && (result = nothing)
     display ≠ Editor() && result ≠ nothing && render(display, result)
     render′(Editor(), result)
   end
@@ -100,7 +100,9 @@ handle("evalrepl") do data
         lock(evallock)
         withpath(nothing) do
           result = @errs eval(mod, :(ans = include_string($code, "console")))
-          !isa(result,EvalError) && ends_with_semicolon(code) && (result = nothing)
+          display = Media.getdisplay(typeof(result), Media.pool(Editor()), default = Console())
+          !isa(result, EvalError) && ends_with_semicolon(code) && (result = nothing)
+          display ≠ Console() && result ≠ nothing && render(display, result)
           render′(result)
         end
         unlock(evallock)


MikeInnes:
This works fine already:

screenshot 2017-07-03 16 01 02

The only reason we need the extra code in `eval` is that we need to choose whether to render twice. But in the base case, Media.jl should just handle this for us.

pfitzseb:
Right, but it might be cool to allow multiple render methods to be called, no? 
E.g. in the case of `Images`, you might want to actually inspect the result of slicing into one in the console instead of in the plotpane (which would have to be the `primarydisplay` to allow rendering the at all).

MikeInnes:
So you think we should render all images in both the console and plot pane?

Should all graphics generally render in both places, as in the editor? This isn't how the repl generally behaves, I guess it could be done.

Bear in mind that this is also user-customisable already; you choose to whether to display images in the plot pane or console even when a default is set.

pfitzseb:
Hm, that's a fair point I suppose.
Still feels kinda weird that I can have something simultaneously displayed in the plot pane and the editor, but not in the console by providing the correct `render` methods.

pfitzseb:
Oh, and getting back to
> Bear in mind that this is also user-customisable already; you choose to whether to display images in the plot pane or console even when a default is set.

How would one do that without knowing the Media internals?

MikeInnes:
To me the editor feels like a special case; ideally, you'd only ever use one display at a time, but it just happens that displaying nothing in the editor is confusing. I think that if we want to support multiple displays generally that needs to become a feature of Media.jl, rather than just an ad-hoc thing per display.

Changing displays should be as simple as `setdisplay(Media.Graphical, Juno.Console())`, but I don't think it's been really well tested by users up to now, so it might not work perfectly.

pfitzseb:
Okay, so this obviously warrants a slightly more nuanced approach than in this PR.

From a user perspective I think it'd be neat to be able to display objects in the plotpane with some general API, just like it's possible to display objects in the console via e.g. `println` or `show`.
What I think might be nice is to have
```julia
img = rand(RGB{N0f8}, 256, 256) # displays via `show` in the console/editor
display(img) # plots in the plot pane; same as `Juno.render(Juno.PlotPane(), img)`
```
but that would probably be abusing Base's display system, and also unnecessarily conflating the Media and Base display systems.

As my confused rambling in this PR indicates, I have no clue how to handle all this consistently though, so I'll think about this some more ;)

MikeInnes:
It is a tricky problem, and I don't think we have the perfect answer yet.

It certainly seems sensible to have a shortcut like `displayimg(x) = render(PlotPane(), x)`. I strongly dislike the current Images approach of overriding display for base arrays, so making that kind of display explicit but convenient seems like a sensible alternative.

pfitzseb:
One thing that just occurred to me is to have a display stack in Media.jl (just like in Base, iirc) and use only the topmost display by default. Then we could also have a `displayall` function that displays the given object in each display (or `displayother`, which displays it in every display except for the default one).

PR #96, opened 2017-07-14
Call invokelatest on `render` and `media`

pfitzseb:
When a package defines a display for a certain type via `Media.media`, that internally calls something like `@eval media(...)`; right after that it's impossible to call any `render` methods, because of world age issues. 
Example would be something like `using Plots; plot([1,2,3])`, which only properly works on the subsequent call to `plot()`.

Together with https://github.com/JuliaPlots/Plots.jl/pull/916 (and https://github.com/JunoLab/atom-ink/pull/139, I guess) this makes the Plots.jl experience much nicer.

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -151,7 +151,7 @@ end
 
 ismacro(f::Function) = startswith(string(methods(f).mt.name), "@")
 
-wstype(x) = nothing
+wstype(x) = ""
 wstype(::Module) = "module"
 wstype(f::Function) = ismacro(f) ? "mixin" : "function"
 wstype(::Type) = "type"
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I noticed this. If you evaluate a bunch of commands all at once and it precompiles, then none of those will render properly. I assume this fixes that?

pfitzseb:
Most likely, yeah. 
Something like `using BenchmarkTools; @benchmark sin(3.2)` works fine with this PR, but spits out `BenchmarkTools.Trial(BenchmarkTools.Parameters(5.0, 10000, 999, 0.0, ...` on Atom master.

MikeInnes:
Makes sense. Would the code be cleaner if we did something like wrapping the whole block in `invokelatest() do ...`?

pfitzseb:
Cleaner maybe, but honestly I'd like to keep the use of `invokelatest` to a minimum. There's probably no good reason for that though, since this code isn't performance critical at all...

MikeInnes:
Surely that would reduce the number of `invokelatest` calls?

pfitzseb:
Sure, but wouldn't you then have some overhead for every single function call in that block? (As you may see from that question, I have no idea how `invokelatest` actually works :D)

MikeInnes:
I don't think so. The overhead should only be at the `invokelatest` call site, but the inner function will be compiled and run as usual. So it would probably be faster to group it together, if anything.

pfitzseb:
Fair enough -- will make the change ;)

PR #98, opened 2017-08-04
reverse backtrace-reversion

pfitzseb:
also better rendering of constructors in backtraces.

Before:
image

After:
image

~~Fixes https://github.com/JunoLab/atom-julia-client/issues/374~~.

### DIFF
--- a/src/display/errors.jl
+++ b/src/display/errors.jl
@@ -57,9 +57,10 @@ end
 function renderbt(trace::StackTrace)
   span(".error-trace",
        [div(".trace-entry $(locationshading(string(frame.file)))",
-            [fade("in "), string(frame.func), fade(" at "),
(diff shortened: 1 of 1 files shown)


MikeInnes:
The current is a better default. Most users are not going to be working on base code, so the user code should come up first.

pfitzseb:
Hm, fair enough. But there's merit to following Base's/the REPL's way of doing things as well:
```
julia> f(0,0)
ERROR: ArgumentError: invalid rational: zero(Int64)//zero(Int64)
Stacktrace:
 [1] Rational{Int64}(::Int64, ::Int64) at ./rational.jl:13
 [2] f(::Int64, ::Int64) at ./REPL[3]:1
```

PR #101, opened 2017-10-31
Stepper

pfitzseb:
ToDo: Lazy loading of ASTInterperter2.jl.


pfitzseb:
Alright, so lazy loading is implemented now and ASTInterperter2.jl is releases. Need to add that dependency here and we should be good to go! :)

MikeInnes:
Great, glad you got the stepper working. I'll play around with this and we can merge it today.

Just cross linking for reference:
https://github.com/JunoLab/atom-julia-client/pull/389
https://github.com/JunoLab/Juno.jl/pull/80
https://github.com/JunoLab/atom-ink/pull/153

MikeInnes:
This is looking really good.

We should clean up the toolbar buttons at some point (probably as part of the terminal work). I think we can pare down the useless buttons, but also give more of juno's tools some visibility.

Since the original version this has lost integration with the working indicator (we had some way of temporarily lowering the count, so that juno isn't constantly marked as working during debug mode). Can you add that back in? If you can't do it soon I'm happy to merge all this as-is.

pfitzseb:
Re buttons: All the debugger buttons are justified imho. Did the "step to selected line" button work well for you?

I added the progress stuff in intentionally, so yes, it's trivial to revert that. might still be nice to have some indication that we're currently in debugger mode.

MikeInnes:
Perhaps; it seems like a bit of an overload to someone who doesn't need it all straight away, but the alternative is probably worse. TBH I was thinking more of the non-debugger stuff anyway.

Yes, skip to line works great, in a quick test at least.

Maybe we should have a debug icon in the status bar or something. It's not a good idea to hide whether or not Juno is actually blocked on something (i had a lot of issues with that in the first version, although ASTInterpreter feels a bit snappier this time around).

PR #106, opened 2018-01-09
Use base display system for plots in repl

pfitzseb:
This allows using Plots.jl (and presumably other plotting packages) in the new REPL-based console together with the plotpane, but unfortunately doesn't yet take plotpane size into account. I'm not sure how to best handle that, but thought I'd throw this out there and hope someone has a good idea.

cc @MikeInnes and @ChrisRackauckas

### DIFF
--- a/src/repl.jl
+++ b/src/repl.jl
@@ -193,26 +193,10 @@ function changeREPLmodule(mod)
   end
 end
 
-struct PlotPaneDisplay <: Base.Multimedia.Display end
-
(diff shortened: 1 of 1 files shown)


MikeInnes:
We should be able to capture display calls via media.jl as usual, no? Blink.jl should actually already provide something similar (i.e. plotting in the terminal opens a blink window).

pfitzseb:
I'm not sure how that would work -- does Media handle mimetypes at all?

PR #114, opened 2018-01-29
reverse stacktraces (again)

### LINKED ISSUE #374: Stack trace reversed in juno
When given a stack trace in the REPL, the innermost call is displayed on top, and the outermost in the bottom. In error pop-ups in Atom/Juno, this order is reversed, with the innermost call in the bottom. It may seem like a small detail, but frequently confuses me when I’m inspecting the traces. 

Brief discussion in https://discourse.julialang.org/t/stack-trace-reversed-in-juno/5199

pfitzseb:
Fixes https://github.com/JunoLab/atom-julia-client/issues/374.

I'm strongly in favour of this just to be consistent -- is this okay with you, @MikeInnes?

### DIFF
--- a/src/display/errors.jl
+++ b/src/display/errors.jl
@@ -64,7 +64,7 @@ function renderbt(trace::StackTrace)
              fade(" at "),
              render(Inline(), Copyable(baselink(string(frame.file), frame.line))),
              fade(frame.inlined ? " <inlined>" : "")])
-        for frame in reverse(trace)])
+        for frame in trace])
 end
 
 rendererr(err) = strong(".error-description", err)


MikeInnes:
Well we could also be consistent, at least within Juno, by changing the REPL stack trace order.

I can go along with this if there's a general consensus that consistency is more important here, but I really do think it's the wrong default for 99% of Julia users (especially the ones using Juno), who usually have an issue in their code, not at the deepest level of the base system.

aviatesk:
I'm actually very in favour of @pfitzseb .
I'm just an inexperienced Julia programmer, but feel like we beginners want to first refer to the outermost call and then continue to dig into inner calls since it would be easier to start thinking from higher level of the error (assuming people usually look at errors from up to down).

PR #134, opened 2018-10-25
render all AbstractDict as a Tree

visr:
I wanted to use nested `OrderedDict`, but was missing the nice drop down display of the `Dict`. I think this same method should be fine for any `AbstractDict` with this minor change.

Before:
image
After:
image


pfitzseb:
Looks good, thanks! :)

PR #136, opened 2018-11-21
legcay -> legacy

tlienart:
I'm assuming that's a typo, otherwise this can be closed.


pfitzseb:
Ha, that is indeed a typo. Thanks for catching it and the PR :)

PR #145, opened 2019-07-19
Add hover

aviatesk:
## Feature - On hover docs

This PR adds "On hover documentation" to Juno.
The PR on atom-julia-client side is: https://github.com/JunoLab/atom-julia-client/pull/573

## Overview

ezgif com-video-to-gif

## How this works

There are Atom packages that offers the Datatip service like above and Juno will consume it. We have two choices for the service package:
- [x] [Atom-IDE-UI](https://github.com/facebookarchive/atom-ide-ui)
    * Pros:
        + Can render the docs very beautifully
        + An user can pin the datatip.
        + Maybe some user will end up enjoying hyperclick feature as well by installing this ? I couldn't find any documentation for it.
    * Cons:
        + Bundles with the other IDE-UI dependencies packages, and it would make the Atom's starting up time more awful
        + Already dead and no future improvements.
- [x] [atom-ide-datatip](https://github.com/atom-ide-community/atom-ide-datatip)
    * Pros:
        + Not bundled with the other packages that aren't be consumed by Juno, and might be better starting up time.
            - Still it needs two dependency packages tho
        + Community-maintained, and hopefully improves in the future ?
    * Cons:
        + Rendering is still not so fancy IMHO
        + Lacks some features that are offered by Atom-IDE-UI:
            - An user can't pin a datatip
            - Code snippet rendering won't work well

## Probs

I'm using two regexs in this PR and they might not be so robust:
- Regex to find code blocks from `string(Markdown.MD)`. Simply don't feel like it's a cool way to do this
https://github.com/JunoLab/Atom.jl/blob/0b2c5950a664f1149d51560d1957e0c2be177b66/src/datatip.jl#L47-L50

- ~~Regex to find if documentation/binding for the word exists. Might not be robust to future changes.~~ 
Seems not so weak (we use the same kind of regex in the other part of code base and we can easily find the change if it happens in the future)

## Announce how to enable the feature ?

And I think we have to announce how to enable this feature somewhere, like on the discourse or documentation.
Anyway I would really appreciate it if you allow me to take an honor to make a new release post on discourse.
Additionally that's would be a good chance to inform of the hyperclick feature again, I think.


src/datatip.jl
```diff
@@ -0,0 +1,91 @@
...
+  push!(datatips, Dict(:type  => :markdown,
+                       :value => header * body))
+end
+
+function makedatatip(docs, word, mtable)
+  # @FIXME?: Separates code blocks from the other markdown texts in order to
+  #          render them as code snippet text by atom-ide-ui's datatip service.
+  #          Setting up functions to deconstruct each `Markdown.MD.content`
+  #          into an appropriate markdown string might be preferred.
+  texts = split(string(docs), codeblock_regex)
```
pfitzseb: Can the datatip package render HTML? If so it might be easiest to use [this](https://github.com/JunoLab/Atom.jl/blob/master/src/display/markdown.jl) and atom-julia-client's `views.render` together (or transform `Hiccup.Node`s into dicts, I guess).

aviatesk: I tried to use them first, but the datatip packs only accept virtual doms made by react...

If we still stick on using our dom, we have to do [something trickly like this](https://github.com/TypeStrong/atom-typescript/blob/master/lib/main/atom-ide/datatipProvider.tsx).

PR #146, opened 2019-07-26
Fix open file

aviatesk:
Improves `openFile` messaging:

- Don't cause error when `InteractiveUtils.edit` gets called with non-Julia files: This is actually a bug
- Now we can open non-Julia files: Why `abspath` is introduced

### DIFF
--- a/src/Atom.jl
+++ b/src/Atom.jl
@@ -25,7 +25,7 @@ function __init__()
               f = Base.find_source_file(path)
               f !== nothing && (path = f)
           end
-          $(msg)("openFile", f, line-1)
+          $(msg)("openFile", Base.abspath(path), line-1)
         end
       end)
     end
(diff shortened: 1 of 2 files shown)


src/Atom.jl
```diff
@@ -25,7 +25,7 @@ function __init__()
               f = Base.find_source_file(path)
               f !== nothing && (path = f)
           end
-          $(msg)("openFile", f, line-1)
+          $(msg)("openFile", Base.abspath(path), line-1)
```
pfitzseb: Did you check that this works fine with remote files?

aviatesk: This is what I should have mentioned in the PR description.

Actually I don't have an easy setup to use remote servers in Juno right now, and didn't check that, sorry.

Maybe could you check this instead ? We may need to use utilities within utils.jl for this to work perfect in every case I suspect.

pfitzseb: Works as expected, afaict.

aviatesk: Thanks for testing ;)

PR #150, opened 2019-08-04
Avoid error on `changeREPLprompt` on REPL startup

aviatesk:
Avoid error:
```
Julia Client – Internal Error

type Nothing has no field current_mode
getproperty(::Any, ::Symbol) at sysimg.jl:18
#changeREPLprompt#167(::Symbol, ::Bool, ::Function, ::String) at repl.jl:111
changeREPLprompt at repl.jl:102 [inlined]
(::getfield(Atom, Symbol("##157#158")))(::String) at repl.jl:23
handlemsg(::String, ::String) at comm.jl:164
(::getfield(Atom, Symbol("##19#21")){Array{Any,1}})() at task.jl:259
```

(Just for those who use the `consoleOptions.terminal` setting)

### DIFF
--- a/src/repl.jl
+++ b/src/repl.jl
@@ -47,7 +47,7 @@ handle("validatepath") do uri
   return isfile′(fullREPLpath(uri)[1])
 end
 
-juliaprompt = "julia> "
+const juliaprompt = "julia> "
 
 handle("resetprompt") do linebreak
   isREPL() || return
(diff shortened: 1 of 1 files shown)


pfitzseb:
Wait, when are you getting that error? IIRC that should only happen when Julia is started with the fallback REPL instead of the `LineEditREPL`, right?

aviatesk:
Can happen an user has settings to activate OhMyREPL in `startup.jl` like:
```julia
atreplinit() do repl
    @time begin
        try @info "Setting up REPL ..."
            @eval using REPL

            if !isdefined(repl, :interface)
                repl.interface = REPL.setup_interface(repl)
            end

            @eval using OhMyREPL
        catch err
            @error err
        end

        try @info "Setting up Revise ..."
            @eval using Revise
            @async Revise.wait_steal_repl_backend()
        catch err
            @error err
        end
    end
end
```

PR #151, opened 2019-08-10
Better completions

aviatesk:
- Ref: https://julialang.slack.com/archives/C7JT7HQAD/p1565114009249700

In short, I want to suppress too many completions from texts like `[|`, `(|`, or `&|` (where `|` represents cursor position), since those over 1000 completions often takes much time and worsen coding experience.

I think it would be better if we could handle these cases in front end (i.e. atom-julia-client) in terms of efficiency, but I found the corner cases can be a lot and handling them in backend like this is fairly easy and simple.


aviatesk:
the latest commit enhances `completionmodule` so that `MethodCompletions` correctly shows the module where the method is defined:
image

aviatesk:
the latest commit enhances completion so that it works in modules that are loaded but not in Main

pfitzseb:
This does feel much better, thanks!

PR #153, opened 2019-08-18
consistent types

aviatesk:
## purpose

addressing: https://github.com/JunoLab/atom-julia-client/blob/master/lib/runtime/completions.js#L4
- we basically use `wstype` and `wsicon` for all the type/icon definitions of workspace/completions/docs
  * simplify code a lot
  * special cases like `keyword` and workspaces in debugger are gracefully handled, I believe
- all these tasks are handled by backend, which makes code simpler I believe
- introduced constant-ness check for all of them: every objects has either `variable` or `constant` type
- added some additional icon definitions for `Regex` and so

Screenshot 2019-08-18 23 55 59

## refs

corresponding PRs:
- https://github.com/JunoLab/atom-julia-client/pull/592
- https://github.com/JunoLab/atom-ink/pull/213

### DIFF
--- a/src/workspace.jl
+++ b/src/workspace.jl
@@ -1,48 +1,46 @@
-ismacro(f::Function) = startswith(string(methods(f).mt.name), "@")
+handle("workspace") do data
+  @destruct [mod, hide || true] = data
+  mod = getmodule′(mod)
+  ns = Symbol.(CodeTools.filtervalid(names(mod; all = true)))
+  filter!(ns) do n
+    (!hide || !Base.isdeprecated(mod, n)) && isdefined(mod, n) && n != Symbol(mod)
+  end
+  contexts = [d(:context => string(mod), :items => map(n -> wsitem(mod, n), ns))]
+  isdebugging() && prepend!(contexts, JunoDebugger.contexts())
+  contexts
+end
 
-wstype(x) = ""
-wstype(::Module) = "module"
-wstype(f::Function) = "function"
-wstype(::Type) = "type"
-wstype(::Expr) = "mixin"
-wstype(::Symbol) = "tag"
-wstype(::AbstractString) = "property"
-wstype(::Number) = "constant"
-wstype(::Exception) = "tag"
-
-wsicon(x) = ""
-wsicon(f::Function) = ismacro(f) ? "icon-mention" : ""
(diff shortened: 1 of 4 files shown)


src/workspace.jl
```diff
@@ -1,48 +1,58 @@
...
+#=
+@NOTE: `wstype` and `wsicon` are also used for completions / docs
+=#
 
-function wsitem(name, val)
-  d(:name  => name,
-    :value => render′(Inline(), val),
-    :type  => wstype(val),
-    :icon  => wsicon(val))
+wstype(mod, name, val) = begin
```
pfitzseb: This seems less clear than the previous dispatch based solution. What's the reason for the change?

aviatesk: I agree with it's less clear and readable.

Why I changed into `isa` (`<:`) branching instead of dispatches is performance: dispatches are slower than explicit conditional branching when the arguments aren't concrete
(You can find my simple benchmarks here: https://gist.github.com/aviatesk/c8c0a24d30051151433f44431e7f54f1)

Of course dispatches automatically do type specializations but I'm fairly sure about the branching above works (even though I suspect we can improve the logic around: https://github.com/JunoLab/Atom.jl/blob/764dfc4aca8cdc61dc35521588cbd7fea3cc5f69/src/workspace.jl#L31-L33 especially)

aviatesk: I.e.: I made a point of not doing something like:
```julia
isabranch() = begin
  if ary isa AbstractArray ...
  elseif ary isa AbstractVector ... # never reached
  else # fallback
  end
end
```

But maybe readability might more important here for us, thus I can change it into dispatch-based branching :)

aviatesk: I just simplified the logic and I hope it makes more sense to you 👍

aviatesk:
included the update to julia 1.2 within completions in the last commit 👍

pfitzseb:
Ok, cool. Feel free to merge when you think it's ready!

aviatesk:
The dispatch vs. isa branching stuff is okay with you ?

pfitzseb:
I'd still like a dispatch based solution better, but it's not a very important change.

aviatesk:
Sure, then I gonna change them into dispatches, actually I like dispatches as an appearance :)

PR #157, opened 2019-08-25
completion test

aviatesk:
I want to add tests for completions since I love that:
- Any error is handled in the parent function `basecompletionadapter`, thus without tests it would need manual testing/detection to find errors when they happen

Would like to merge https://github.com/JunoLab/Atom.jl/pull/156 first


aviatesk:
will add some more tests

pfitzseb:
Nice, good work!

PR #158, opened 2019-08-26
more package completion

aviatesk:
Packages that aren't loaded in the `load_path` can't be completed, since 
- it's returned as `ModuleCompletios`:
https://github.com/JuliaLang/julia/blob/master/stdlib/REPL/src/REPLCompletions.jl#L644-L647
- and then it's suppressed at:
https://github.com/JunoLab/Atom.jl/blob/master/src/completions.jl#L37

and this allows completions for packages that are only loaded on each module level


aviatesk:
aah, then the completions can be duplicated, since `Main` and the other each module can use the same module ...
I turned on `similarSuggestionRemoval: "textOrSnippet"` config of auto-complete package and never noticed this, but this gonna be pretty annoying.

aviatesk:
Feels like this can appear complicated, thus close this PR.

PR #161, opened 2019-08-31
test, test, and test

aviatesk:
add tests for various things I want to make sure to work:

I also may want to enable codecov for Atom.jl, which may or may not encourage us to write more tests in the future, what do you think @pfitzseb ?

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -56,11 +56,11 @@ function completion(mod, line, c)
               :description => completionsummary(mod, c))
 end
 
-completiontext(x) = completion_text(x)
-completiontext(x::REPLCompletions.PathCompletion) = rstrip(completion_text(x), '"')
-completiontext(x::REPLCompletions.DictCompletion) = rstrip(completion_text(x), [']', '"'])
-completiontext(x::REPLCompletions.MethodCompletion) = begin
-  ct = completion_text(x)
+completiontext(c) = completion_text(c)
+completiontext(c::REPLCompletions.PathCompletion) = rstrip(completion_text(c), '"')
+completiontext(c::REPLCompletions.DictCompletion) = rstrip(completion_text(c), [']', '"'])
+completiontext(c::REPLCompletions.MethodCompletion) = begin
+  ct = completion_text(c)
   ct = match(r"^(.*) in .*$", ct)
   ct isa Nothing ? ct : ct[1]
 end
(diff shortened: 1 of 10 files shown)


aviatesk:
Squashed commits:

refactor tests:
- move basic message handling tests into test/mics.jl
- move test for toggle docs/methods into test/eval.jl
- add tests for workspace
- refactor tests for completions
- add test for field completions

more refactors on completions/workspace
- suppress submodules to appear themselves in their workspace view
- prefer dispatch way for `completiontype`
- consistent method signature for `completiontext`
- no specialize on the argument of wsicon/wstype/completiontype
- suppress misleading descriptions for property/field completions
- enhance return type of field completion

*** 

Still want your review for few points.

src/completions.jl
```diff
@@ -164,16 +170,18 @@ completiontype(line, c, mod) = begin
...
+  completiontype(c)
+end
+
+completiontype(line, ::REPLCompletions.DictCompletion, mod) = "key" # DictCompletion isn't dispatched for the entry method, otherwise fallen into "macro"
+completiontype(@nospecialize(val), mod::Module, ct::String) = wstype(mod, Symbol(ct), val) # ModuleCompletion
+completiontype(::REPLCompletions.Completion) = "variable" # may not happen
+completiontype(::REPLCompletions.PackageCompletion) = "import"
+completiontype(::REPLCompletions.MethodCompletion) = "method"
+completiontype(::REPLCompletions.PropertyCompletion) = "property"
+completiontype(::REPLCompletions.FieldCompletion) = "property"
```
aviatesk: What `FieldCompletion` does is essentially completing properties of a predicted return type, thus I want to change its type into `property`, and keep `attribute` for the future scope-based completions [described here](https://github.com/JunoLab/Atom.jl/blob/master/src/completions.jl#L34) (, which I would like to tackle in the next JSOC or some other place)

src/workspace.jl
```diff
@@ -41,7 +42,7 @@ wstype(mod, name, ::Symbol) = "tag"
 wstype(mod, name, ::Exception) = "mixin"
 wstype(mod, name, ::Undefined) = "ignored"
 
-wsicon(mod, name, val) = isconst(mod, name) ? "c" : "v"
+wsicon(mod, name, @nospecialize(val)) = isconst(mod, name) ? "c" : "v"
```
aviatesk: I believe those uses of `@nospecialize` would be good here, but still want you to confirm that.

pfitzseb: Yeah, makes sense to have this.

src/completions.jl
pfitzseb: Sure, I don't particularly care about the exact symbol we show :)

src/workspace.jl
aviatesk: Thanks for checking this !

PR #162, opened 2019-09-01
Enforce compatibility with julia 1

timholy:
It's probably better not to promise compatibility with Julia 2+

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -45,7 +45,7 @@ Lazy = "^0.13"
 MacroTools = "^0.5"
 TreeViews = "^0.3.0"
 WebIO = "^0.8.1"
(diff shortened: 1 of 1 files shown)


timholy:
Also, we need a new release (see https://discourse.julialang.org/t/juno-support-for-julia-1-2-0-not-ready-yet/27804). Check out the [registry Compat.toml](https://github.com/JuliaRegistries/General/blob/master/A/Atom/Compat.toml), there's no release supporting JuliaInterpreter 0.7.

pfitzseb:
Yeah, I'll tag a new JuliaInterpreter 0.7 compatible release today or tomorrow.

PR #164, opened 2019-09-02
add uri-handler for completions

aviatesk:
can be useful for someone ?

xref: https://github.com/JunoLab/atom-julia-client/pull/601


aviatesk:
this handler can be a temporal solution for link handling in datatips, and gonna implement that.

PR #165, opened 2019-09-02
local completions and outline

pfitzseb:


src/completions.jl
```diff
@@ -42,6 +48,20 @@ function basecompletionadapter(line, mod, force)
       continue
     end
   end
+  
+  # completions from the local code block:
+  for c in locals(text, lineNumber, column)
+    if startswith(c, pre)
```
aviatesk: Seems better to special case `pre == ""`, since otherwise `PathCompletions` get suppressed:
image

src/completions.jl
```diff
@@ -42,6 +48,20 @@ function basecompletionadapter(line, mod, force)
...
+      pushfirst!(d, Dict(
+                  :type        => "",
+                  :icon        => "l",
```
aviatesk: How about this iconography ?:
image

src/outline.jl
```diff
@@ -0,0 +1,97 @@
...
+    byteoffset = 1
+    current_line = 1
+    current_char = 0
+    for c in text
+        if line == current_line
+            current_char += 1
+            c === '\n' && break
+        end
+        current_char == col && break
+        byteoffset += ncodeunits(c)
```
aviatesk: `byteoffset += VERSION >= v"1.1" ? ncodeunits(c) : ncodeunits(string(c))` will fix the problem in Julia 1.0 ?

src/completions.jl
pfitzseb: Good call.

src/completions.jl
pfitzseb: I went with `icon-chevron-right` instead, didn't like the magnifying glass :D

src/outline.jl
pfitzseb: Ah, right. Fixed.

src/completions.jl
aviatesk: Okay !

aviatesk:
Would it be more fancy to search for the root parent `EXPR` for each binding, and shows the parent's binding name in `rightLabel` ?

test/completions.jl
```diff
@@ -113,3 +113,75 @@
         @test length(readmsg()[3]["completions"]) > Atom.MAX_COMPLETIONS # invoked
     end
 end
+
+@testset "local completions" begin
```
aviatesk: Love to have tests for them !!!

pfitzseb:
Hm, maybe. I've implemented that for now:
image

aviatesk:
Seems cool to me 😄

PR #167, opened 2019-09-03
debugger enhancements

aviatesk:
- add different prompt color for compiled mode
- enable basic REPL completions in debugger REPL

we may enhance more, by enabling local completions as well (using the same logic as workspace)

***

additionally, I would like to add debugger commands that are accessible from debugger REPL like [Debugger.jl's interface](https://github.com/JuliaDebug/Debugger.jl#debugger-commands), e.g.:
- entering `:nc` in the REPL will step into the next call.

I would like to hear your ideas about these commands.

### DIFF
--- a/src/debugger/breakpoints.jl
+++ b/src/debugger/breakpoints.jl
@@ -120,12 +120,14 @@ handle("getBreakpoints") do
   end
 end
 
+isCompileMode() = _compiledMode[] === JuliaInterpreter.Compiled()
+
 handle("toggleCompiled") do
   with_error_message() do
-    _compiledMode[] = (_compiledMode[] === JuliaInterpreter.finish_and_return! ?
-      (_compiledMode[] = JuliaInterpreter.Compiled()) :
-      (_compiledMode[] = JuliaInterpreter.finish_and_return!))
-    return _compiledMode[] === JuliaInterpreter.Compiled()
+    _compiledMode[] = (isCompileMode() ?
+      (_compiledMode[] = JuliaInterpreter.finish_and_return!) :
+      (_compiledMode[] = JuliaInterpreter.Compiled()))
+    return isCompileMode()
   end
 end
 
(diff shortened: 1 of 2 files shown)


src/debugger/breakpoints.jl
```diff
@@ -120,12 +120,14 @@ handle("getBreakpoints") do
...
 handle("toggleCompiled") do
   with_error_message() do
-    _compiledMode[] = (_compiledMode[] === JuliaInterpreter.finish_and_return! ?
-      (_compiledMode[] = JuliaInterpreter.Compiled()) :
-      (_compiledMode[] = JuliaInterpreter.finish_and_return!))
-    return _compiledMode[] === JuliaInterpreter.Compiled()
+    _compiledMode[] = (isCompileMode() ?
+      (_compiledMode[] = JuliaInterpreter.finish_and_return!) :
+      (_compiledMode[] = JuliaInterpreter.Compiled()))
+    return isCompileMode()
```
aviatesk: Still now sure why this part wasn't like
```julia
_compiledMode[] = isCompileMode() ? JuliaInterpreter.finish_and_return! : JuliaInterpreter.Compiled()
```

pfitzseb: Probably because I refactored this at some point and never thought about what I wrote there. Definitely makes more sense like this though ;)

pfitzseb:
> additionally, I would like to add debugger commands that are accessible from debugger REPL like Debugger.jl's interface, e.g.:

>    entering :nc in the REPL will step into the next call.

Hm, not sure. Why is this better than just having global keybindings?

aviatesk:
Just because of my preference ? When in debugging and playing around in REPL, I often feel like to control the debug state directly in REPL, with minimum stroke.

src/debugger/breakpoints.jl
aviatesk: fixed ;)

aviatesk:
will add local completion using workspace's `localvars` logic later.

pfitzseb:
Nice, local completions are great to have! Will merge this now, let's move the REPL command discussion to a new issue/PR.

aviatesk:
> Nice, local completions are great to have! Will merge this now, let's move the REPL command discussion to a new issue/PR.

yeah, that's would be better (I tend to do different things in a single PR ... xD)

PR #170, opened 2019-09-09
improve completions/outline

pfitzseb:
- don't include function arguments in outline
- stop kwargs/anon function args form leaking into outer scope
- improve locality based completion ranking by basing it on lines 
instead of the byteoffset
- give bindings defined *before* the current line a higher priority

### DIFF
--- a/src/outline.jl
+++ b/src/outline.jl
@@ -83,19 +83,19 @@ function toplevel_bindings(expr, text, bindings = [], line = 1, pos = 1)
     return bindings
 end
 
-struct LocalScope
-    name
-    span
-    line
-    children
-    expr
+struct LocalBinding
+    name::String
+    span::UnitRange{Int64}
+    line::Int
+    expr::CSTParser.EXPR
 end
 
-struct LocalBinding
-    name
-    span
-    line
-    expr
+struct LocalScope
+    name::String
+    span::UnitRange{Int64}
+    line::Int
+    children::Vector{Union{LocalBinding, LocalScope}}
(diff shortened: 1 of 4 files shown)


src/outline.jl
```diff
@@ -173,6 +218,12 @@ function str_value(x)
         x.kind == CSTParser.Tokens.AT_SIGN && return "@"
         x.kind == CSTParser.Tokens.DOT && return "."
         return ""
+    elseif x.kind === CSTParser.Tokens.TRIPLE_STRING
+        return string("\"\"\"", x.val, "\"\"\"")
+    elseif x.kind === CSTParser.Tokens.STRING
+        return string("\"", x.val, "\"")
+    elseif x.kind === CSTParser.Tokens.CHAR
+        return string("\'", x.val, "\'")
```
aviatesk: I believe you don't need this check for `CSTParser.Tokens.CHAR` or just return `x.val`, since `x.val` already contains `'` as is.

src/outline.jl
```diff
@@ -77,41 +87,63 @@ function toplevel_bindings(expr, text, bindings = [], line = 1, pos = 1)
     return bindings
 end
 
+struct LocalScope
+    name
```
aviatesk: sorry for the annoying rebasing... I just believed the prototyping had finished

src/outline.jl
pfitzseb: Indeed. I only checked the two string cases and assumed chars are handled the same...

PR #173, opened 2019-09-13
show value string within datatip

aviatesk:
- for toplevel bindings
- for local bindings within debugging

e.g.: debugging:
image

ref: https://github.com/JunoLab/atom-julia-client/pull/614

***

I may want to add tests for this, but I suspect the code would change significantly once we move to our own datatip UI and honestly wondering writing tests for now is worth to do.


src/datatip.jl
```diff
@@ -1,46 +1,62 @@
...
 
-  Dict(:error   => false,
-       :strings => makedatatip(docs))
+  # don't show verbose stuff
+  docstr = replace(string(docs), nodoc_regex => "")
+  occursin(nobinding_regex, docstr) && return Dict(:error => true)
+
+  datatip = []
+
+  val = getfield′′(getmodule′(mod), Symbol(word))
```
aviatesk: maybe this line should be replaced with another better way, or needs some tweaks, I think.

src/utils.jl
```diff
@@ -100,6 +100,6 @@ function md_hlines(md)
 end
 
 function strlimit(str::AbstractString, limit = 30)
-  str = lastindex(str) > limit ?  str[1:prevind(str, limit)]*"…" : str
+  str = lastindex(str) > limit ? str[1:prevind(str, limit)]*" ..." : str
```
aviatesk: ` ...` would be more consistent with the same stuff from the usual Atom packages, like autocomplete-plus and so.

src/datatip.jl
aviatesk: should be improved by #174

aviatesk:
I would like to merge this after #174 gets merged.

src/utils.jl
pfitzseb: Honestly, I prefer `…` here -- this should guarantee that your string is *never* longer than `limit`.

aviatesk: ah yeah, I understand.

Apart from that, I think this function is better to respect the _length_ of `str` instead of code units of that ?

I mean the implementation below would behave more intuitively imho:
```jl
function strlimit(str::AbstractString, limit::Int = 30, ellipsis::AbstractString = " ...")
  charary = collect(str)
  str = length(charary) > limit ? join(charary[1:limit - length(ellipsis)])*ellipsis : str
  filter(isvalid, str)
end
```

then `strlimit("jμλια", 5) == "jμλια"` (the current implementation would return `jμλ…` instead).

pfitzseb: Ok, then let's go with 
```
function strlimit(str::AbstractString, limit::Int = 30, ellipsis::AbstractString = "…")
  io = IOBuffer()
  i = 0
  for c in str
    i > limit - length(ellipsis) && break
    isvalid(c) || continue
    
    print(io, c)
    i += 1
  end
  length(str) > limit && print(io, ellipsis)
  
  return String(take!(io))
end
```
instead -- that should be correct in all cases and also have better performance than the array manipulations.

aviatesk: ah yeah, it should be faster, thanks for the suggestion :)

fixed with 51fb20d ;)

pfitzseb: The problem with the `enumerate` solution is that you might discard too many chars:
```
julia> length(strlimit2(String(rand(UInt8, 100)), 20)) # this is my implementation
20

julia> length(strlimit(String(rand(UInt8, 100)), 20))
13
```

My version still isn't entirely correct, so let's use
```
function strlimit2(str::AbstractString, limit::Int = 30, ellipsis::AbstractString = "…")
  will_append = length(str) > limit
  
  io = IOBuffer()
  i = 1
  for c in str
    will_append && i > limit - length(ellipsis) && break
    isvalid(c) || continue
    
    print(io, c)
    i += 1
  end
  will_append && print(io, ellipsis)
  
  return String(take!(io))
end
```
instead :) Would also be great if you could add a test or two, but that doesn't need to hold up this PR.

aviatesk: Aaah, ashamed, thanks for the fix again, now I see why `i` is needed ...

***

> Would also be great if you could add a test or two, but that doesn't need to hold up this PR.
Yeah, would be great. I may try to add tests, even tho I'm not sure I can come up with good ones, but the one using `rand` seems cool to me.

pfitzseb: > Aaah, ashamed, thanks for the fix again, now I see why i is needed ...

No worries, I had your exact version at first too.

> Yeah, would be great. I may try to add tests, even tho I'm not sure I can come up with good ones, but the one using rand seems cool to me.

Maybe also something simple like 
```
@test strlimit("jμλια", 5) == "jμλια"
@test strlimit("jμλια", 4) == "jμλ…"
@test strlimit("abcde", 5) == "abcde"
@test strlimit("abcde", 4) == "abc…"
@test strlimit("jμ"*String(b"\xb6")*"iλια", 5) == "jμλια" # this probably doesn't pass though
```
etc

aviatesk: Hope these are addressed within the latest commit ;)

pfitzseb:
Great work, thanks!

PR #174, opened 2019-09-13
refactors around `get` methods

aviatesk:
refactor get moethods: …
- define getter methods in one place: utils.jl
- add type annotation and docs for better development
- `getfield′` now uses `CodeTools.getthing`
     * e.g.: #173 will be able to handle `Atom.handlers` in `Atom` module

getmodule′ -> getmodule:

### DIFF
--- a/src/Atom.jl
+++ b/src/Atom.jl
@@ -38,6 +38,7 @@ function __init__()
 end
 
 include("comm.jl")
+include("utils.jl")
 include("display/display.jl")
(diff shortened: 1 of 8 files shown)


pfitzseb:
Alright, tests pass so this seems to still do the right thing. Merging :)

aviatesk:
@pfitzseb thanks, and yeah, this PR can appear such a dangerous one, but I believe this still doesn't break anything

PR #175, opened 2019-09-15
reverse workspace stack in debugging

aviatesk:
imho this would make workspace view in debugging more intuitive
- the more recent values are (first) interpreted, the upper they would appear:
- consistent with how the function callstacks are ordered

### DIFF
--- a/src/debugger/stepper.jl
+++ b/src/debugger/stepper.jl
@@ -423,11 +423,11 @@ function contexts(s::DebuggerState = STATE)
     trace = string(trace, "/", frame.framecode.scope isa Method ?
                                 frame.framecode.scope.name : "???")
     c = Dict(:context => string("Debug: ", trace), :items => localvars(frame))
-    push!(ctx, c)
+    pushfirst!(ctx, c)
 
     frame = frame.callee
   end
-  reverse(ctx)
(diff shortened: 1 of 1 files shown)


pfitzseb:
Hm, I don't think I like this all that much -- especially since the ordering of `locals` isn't *really* dependent on where the variable has been defined. The `pushfirst`/`reverse` change is obviously good though :)

aviatesk:
> Hm, I don't think I like this all that much -- especially since the ordering of locals isn't really dependent on where the variable has been defined.
Oh really.. I just took a look at the implementation of `JuliaInterpreter.locals` again, and still think the locals are ordered by their definition.
Maybe you meant `sparams` stuff by "really" ? (just curious)

Anyway, I just reverted the reversing stack, and will merge this for now.

pfitzseb:
>  Maybe you meant sparams stuff by "really" ? (just curious)

Yup. On second thought that's not my actual reason for disliking the change (since the ordering *is* deterministic) -- I just like having the variables show up in the order of definition :P

aviatesk:
> I just like having the variables show up in the order of definition :P

Okay, then let's go with the current behaviour ;)

PR #178, opened 2019-09-15
aggregate methods with default args within goto

aviatesk:
A method definition with default arguments introduces different method instances, but they are better to be handled as an unique method in terms of "goto method", since their definition location is exactly same, and otherwise they might appear verbose.

- the methods with full arguments are going to be shown as a representative of the other _incomplete_ methods
- Tests are added.


pfitzseb:
Nice, I like it.

PR #180, opened 2019-09-17
enhance completions within the integrated REPL

aviatesk:
First I tried to enable this feature via https://github.com/JuliaLang/julia/pull/33102, but the PR would take more time to be finished I think, and so I want to _partly_ realise that here:

Pros:
- we can get module-aware completions within the integrated REPL

Cons:
- an user can no longer has their own custom `complete` within Juno: it's automatically updated by each call of `changemodule`
- help mode is still not module-aware (that's why this feature is still "partly" realised)

Alternative design:
- if we use an weird global `context_module` for REPL completions, then we can attach our custom completion provider on `atreplinit`, and then it would allow an user's custom `complete`

### DIFF
--- a/src/repl.jl
+++ b/src/repl.jl
@@ -14,9 +14,17 @@ function get_main_mode()
   mode
 end
 
-isREPL() = isdefined(Base, :active_repl) &&
-           isdefined(Base.active_repl, :interface) &&
-           isdefined(Base.active_repl.interface, :modes)
+function isREPL(; before_run_repl = false)
+  isdefined(Base, :active_repl) &&
+  isdefined(Base.active_repl, :interface) &&
+  isdefined(Base.active_repl.interface, :modes) &&
+  if before_run_repl
+    true
+  else
+    isdefined(Base.active_repl, :mistate) &&
+    isa(Base.active_repl.mistate, REPL.LineEdit.MIState)
+  end
+end
 
 handle("changeprompt") do prompt
   isREPL() || return
@@ -61,7 +69,7 @@ end
 const current_prompt = Ref{String}(juliaprompt)
 
 function hideprompt(f)
-  (isREPL() && Base.active_repl.mistate isa REPL.LineEdit.MIState) || return f()
+  isREPL() || return f()
 
   repl = Base.active_repl
   mistate = repl.mistate
@@ -118,7 +126,7 @@ function changeREPLprompt(prompt; color = :green, write = true)
     print(stdout, "\e[1K\r")
     REPL.LineEdit.write_prompt(stdout, main_mode)
   end
-  global current_prompt[] = prompt
+  current_prompt[] = prompt
   nothing
 end
 
--- a/src/repl.jl
+++ b/src/repl.jl
@@ -189,6 +189,7 @@ function changeREPLmodule(mod)
 
   mod = getmodule(mod)
 
+  # change the repl module in the context of both evaluation and completions
   main_mode = get_main_mode()
   main_mode.on_done = REPL.respond(Base.active_repl, main_mode; pass_empty = false) do line
     if !isempty(line)
@@ -197,6 +198,7 @@ function changeREPLmodule(mod)
       end
     end
   end
+  main_mode.complete = JunoREPLCompletionProvider(mod)
 end
 
 function reset_repl_history()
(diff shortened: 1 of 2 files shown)


src/repl.jl
```diff
@@ -58,10 +58,10 @@ handle("resetprompt") do linebreak
   nothing
 end
 
-current_prompt = juliaprompt
+const current_prompt = Ref{String}(juliaprompt)
 
 function hideprompt(f)
-  isREPL() || return f()
+  (isREPL() && Base.active_repl.mistate isa REPL.LineEdit.MIState) || return f()
```
aviatesk: this change is because I encountered an error that says `mistate` is still `nothing` if we start Julia via command like `julia-client: run-all`.

I suspect this happens `hideprompt` can be called before `active_repl.interface` is fully setup:
- I still not tried to fix our script loading/activation order, but I think adding the snippet at our `atreplinit` may solve this all ?
```jl
if !isdefined(repl, :interface)
    repl.interface = REPL.setup_interface(repl)
end
```

src/repl.jl
```diff
@@ -235,3 +237,20 @@ function fixjunodisplays()
...
+  mod::Module
+end
+
+function LineEdit.complete_line(c::JunoREPLCompletionProvider, s)
+  partial = REPL.beforecursor(s.input_buffer)
+  full = LineEdit.input_string(s)
+
+  # module-aware repl backend completions
+  comps, range, should_complete = REPLCompletions.completions(full, lastindex(partial), c.mod)
+  ret = map(REPLCompletions.completion_text, comps) |> unique!
```
aviatesk: I may want to show `ModuleCompletion`s with its module identical to `JunoREPLCompletionProvider.mod` up: this won't be so computational heavy but what do you think about this idea ?

src/repl.jl
pfitzseb: Just put that check into `isREPL` -- we're assuming that that's true all over the place anyways.

src/repl.jl
pfitzseb: Sorry, not sure what you mean.

aviatesk: NVM... I had a kinda same idea within debugger local completions, but now I find there is no need for this here at all.

src/repl.jl
```diff
@@ -118,7 +118,7 @@ function changeREPLprompt(prompt; color = :green, write = true)
     print(stdout, "\e[1K\r")
     REPL.LineEdit.write_prompt(stdout, main_mode)
   end
-  global current_prompt = prompt
+  global current_prompt[] = prompt
```
pfitzseb: Nitpick, but we don't need the `global` here.

src/repl.jl
aviatesk: I didn't do that since then [this part](https://github.com/JunoLab/Atom.jl/blob/master/src/Atom.jl#L17-L33) wouldn't get fired since it's supposed to be called before `run_main_repl` is called: https://github.com/JuliaLang/julia/blob/master/base/client.jl#L312

I special cased it in the last commit, and does it sound fine ?

src/repl.jl
```diff
@@ -14,9 +14,17 @@ function get_main_mode()
...
 end
 
-isREPL() = isdefined(Base, :active_repl) &&
-           isdefined(Base.active_repl, :interface) &&
-           isdefined(Base.active_repl.interface, :modes)
+function isREPL(; before_run_repl = false)
+  isdefined(Base, :active_repl) &&
+  isdefined(Base.active_repl, :interface) &&
+  isdefined(Base.active_repl.interface, :modes) &&
+  if before_run_repl
```
pfitzseb: Why not just check that all the time? Both the `isdefined` and the `isa` call should be very cheap.

aviatesk: This comment explains it ?
https://github.com/JunoLab/Atom.jl/pull/180#discussion_r325072523

src/repl.jl
aviatesk: `run_main_repl` sets up `active_repl.mistate` but [`connected`](https://github.com/JunoLab/Atom.jl/blob/master/src/Atom.jl#L16) can be called before `run_main_repl`, can't it ?

aviatesk:
@pfitzseb 

> Cons:
> - an user can no longer has their own custom complete within Juno: it's automatically updated by each call of changemodule

Not a problem ?

pfitzseb:
Do you mean that users can't overwrite `Base.active_repl.interface.modes{1].complete` anymore because we would nix their change on the next call to `changeREPLmodule`? If so then yes, that's not a problem -- they shouldn't really be doing that anyways :P

aviatesk:
> Do you mean that users can't overwrite Base.active_repl.interface.modes{1].complete anymore because we would nix their change on the next call to changeREPLmodule?

Yes, exactly. Then the current implementation would be acceptable.

I’m just curious why these module-aware completions within REPL are not implemented until now, and wondered my implementation has some serious trouble or not.

pfitzseb:
I think the last time I looked at it `REPLCompletions.completions` didn't take the `mod` argument. 
AFAICT your implementation should work just fine.

aviatesk:
I see. Thanks for your review !

PR #181, opened 2019-09-17
debugger refactors

aviatesk:
Mainly I did two things on this branch:
1. separate stepper.jl into appropriate sub files (since stepper.jl's got so fat)
2. enable "complete" completions within debug REPL: e.g.:
    * `MethodCompletion` including local bindings: `strlimit(localstring, |` should show `MethodCompletion`s for `strlimit(str::String, ...)`
    * `PropertyCompletion` for local bindings: `localobj.|` should show the properties of `localobj`
    * etc...

A part that needs to be reviewed is obviously around 2 -- it's doing [such a hacky/dirty thing](https://github.com/JunoLab/Atom.jl/blob/01c215ba246e9c0d3a159b045b829f0533d8a9bf/src/debugger/debugger.jl#L26-L67) (overwriting the methods within REPL module)
This implementation is not so robust to future changes within `REPLCompletion` module, but imho I suspect we can detect future regressions easily, since now we have kinda rich tests for that. Still you may think this implementation is not good in terms of maintenance though.

As for the refactors around 1, I just move the code and removed unnecessary imports carefully.
 
***

UPDATE:

This PR only does several refactors around `Atom.JunoDebugger`, and the hacks on `REPLCompletions` are discarded.
If anyone interested in that, visit [my startup script](https://github.com/aviatesk/avi-atom/blob/master/scripts/junostartup.jl#L45-L83)

Refactors:
1. separate stepper.jl into appropriate sub files (since stepper.jl's got so fat)
2. suppress constantness iconography within debug workspace items: Julia doesn't support explicit constantness for local variables
3. send `doneWorking` message if a sent line is empty within debug repl
4. make debug datatip works for base funcitons

### DIFF
--- a/src/debugger/datatip.jl
+++ b/src/debugger/datatip.jl
@@ -3,11 +3,14 @@ using JuliaInterpreter: getfile, locals
 import ..Atom: fullpath, strlimit
 
 function datatip(word, path, row, column, state::DebuggerState = STATE)
-  frame = state.frame
+  # frame validation
+  state.frame === nothing && return nothing
+
+  frame = active_frame(state)
   scope = frame.framecode.scope
 
-  # frame validation & only work for methods
-  (frame === nothing || scope isa Module) && return nothing
+  # only work for methods
+  scope isa Module && return nothing
 
   # path identity check
   path != fullpath(getfile(frame)) && return nothing
--- a/src/debugger/datatip.jl
+++ b/src/debugger/datatip.jl
@@ -1,6 +1,6 @@
 using CodeTracking
 using JuliaInterpreter: getfile, locals
-import ..Atom: strlimit
+import ..Atom: fullpath, strlimit
 
 function datatip(word, path, row, column, state::DebuggerState = STATE)
   frame = state.frame
@@ -10,7 +10,7 @@ function datatip(word, path, row, column, state::DebuggerState = STATE)
   (frame === nothing || scope isa Module) && return nothing
 
   # path identity check
-  path != getfile(frame) && return nothing
+  path != fullpath(getfile(frame)) && return nothing
 
   # line number check
   defstr, startline = definition(String, scope)
--- a/src/debugger/datatip.jl
+++ b/src/debugger/datatip.jl
@@ -2,14 +2,12 @@ using CodeTracking
 using JuliaInterpreter: getfile, locals
 import ..Atom: strlimit
 
-function datatip(word, path, row, column, s::DebuggerState = STATE)
-  frame = s.frame
+function datatip(word, path, row, column, state::DebuggerState = STATE)
+  frame = state.frame
+  scope = frame.framecode.scope
 
   # frame validation & only work for methods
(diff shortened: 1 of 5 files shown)


aviatesk:
Refactors added:
- keep local variables for the current `STATE.frame` in `STATE`: avoids duplicated computation of them within REPL completions/datatip
- suppress constantness iconography within workspace: Julia doesn't support explicit constantness for local variables

pfitzseb:
I'm a big fan of the refactoring, but would actually prefer not hacking in more completions. Overwriting `Base` internals always feels so ... wrong.

aviatesk:
> Overwriting Base internals always feels so ... wrong.

I understand, sincerely, haha. It's definitely not robust (the hack is even relying on the global `STATE`, which would also be _wrong_)
It's kinda too dangerous to be published and I'm really okay just to keep this hack on my startup script, and I just wanted to show the possibility of this ?

***

Okay, then I'm going to revert the hacks on completions and keep only the refactors :)

aviatesk:
Manually reverted 👍

src/debugger/repl.jl
```diff
@@ -0,0 +1,94 @@
...
+    end
+  end) map(v -> string(v.name)) prepend!(ret) unique!
+
+  return ret, partial[range], should_complete
+end
+
+# Evaluation
+function interpret(code::AbstractString, state::DebuggerState = STATE)
+  state.frame === nothing && return
+  eval_code(active_frame(state), code)
```
aviatesk: I tried to understand the behaviour of `active_frame` and `STATE.level`, but I'm still not sure why `active_frame` is needed.

Is there a case when `STATE.frame` doesn't point to the innermost (i.e. active) frame ?
Should I need to use this logic in datatip as well ?

src/debugger/datatip.jl
```diff
@@ -0,0 +1,29 @@
...
+
+function datatip(word, path, row, column, state::DebuggerState = STATE)
+  frame = state.frame
+  scope = frame.framecode.scope
+
+  # frame validation & only work for methods
+  (frame === nothing || scope isa Module) && return nothing
+
+  # path identity check
+  path != fullpath(getfile(frame)) && return nothing
```
aviatesk: I think this would work for remote files, but haven't tried that actually.

src/debugger/repl.jl
pfitzseb: So `STATE.frame` is the frame we stopped interpreting in, but you can also switch to a parent frame by clicking on the stackframe number (which is saved in `STATE.level`). Code is then evaluated in that context. So yes, the debugging datatips need `active_frame(state)` instead of `state.frame` as well.

aviatesk: I see, I missed to check the `setStackLevel` handler.

So then maybe using `active_frame` in
- repl completions
- workspace (only show local binds up to the `active_frame(state)`)
- datatip

would be more consistent I think ?

(I included the change in f188e4c)

PR #182, opened 2019-09-21
different vector/set/dict head render

pfitzseb:
image

Fixes https://github.com/JunoLab/Juno.jl/issues/371.

### DIFF
--- a/src/display/base.jl
+++ b/src/display/base.jl
@@ -112,11 +112,11 @@ trim(xs, len = 25) =
                     [undefs(xs[1:10]); fade("..."); undefs(xs[end-9:end])]
 
 @render i::Inline xs::Vector begin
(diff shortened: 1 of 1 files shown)


src/display/base.jl
```diff
@@ -134,8 +134,8 @@ end
     end
     return st
   end
-  LazyTree(span(c(strong(String(nameof(typeof(d)))),
-            Atom.fade(" $(eltype(d).parameters[1]) → $(eltype(d).parameters[2]) with $(length(d)) entries"))), cs)
```
aviatesk: how about changing `→`s on line 131 & 133 as well ?
It would make it more consistent with the syntax (and the `display`)

image

pfitzseb: Yeah, makes sense, I suppose.

JeffBezanson:
Beautiful, thanks!

PR #184, opened 2019-09-24
stringify method signatures with keyword args correctly

aviatesk:
fixes https://github.com/JunoLab/Juno.jl/issues/374

### DIFF
--- a/src/outline.jl
+++ b/src/outline.jl
@@ -93,9 +93,9 @@ function toplevel_bindings(expr, text, bindings = [], line = 1, pos = 1)
 end
 
 function Base.countlines(expr::CSTParser.EXPR, text::String, pos::Int, full::Bool = true; eol = '\n')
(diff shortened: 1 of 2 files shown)


pfitzseb:
Needs a rebase.

src/outline.jl
```diff
@@ -245,13 +245,11 @@ function str_value(x)
...
         return CSTParser.str_value(x)
     else
-        s = ""
-        for a in x
-            s *= str_value(a)
-        end
-        return s
+        return join(str_value(a) for a in x)
     end
 end
```
aviatesk: @pfitzseb do you mind moving the definition of `str_value` to after `toplevel_binding` ? I want keep related code aggregated if possible.

aviatesk: NVM, should be done in another PR.

PR #185, opened 2019-09-26
more refactors on completions.jl

aviatesk:
- simplify the whole logic in `completion(mod, c)`
- `returntype` -> `completionreturntype`
- enhance `DictCompletion`
  * show value type
  * add icon/type
- enhance `FieldCompletion`
  * show a predicted type in rightLabel
- fix the logic `completiontext(::MethodCompletion)` for a case when the regex does not match
- `shortstr` function
- move `cangetdocs` into utils.jl
- don't use `handle` in tests
- add more test cases

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -42,7 +42,7 @@ function basecompletionadapter(line, mod, force, lineNumber, column, text)
       c isa REPLCompletions.PackageCompletion || continue
     end
     try
-      push!(d, completion(mod, line, c))
+      push!(d, completion(mod, c))
     catch err
       continue
     end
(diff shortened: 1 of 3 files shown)


aviatesk:
The test fails on nightly, but it's not related to the changes made in this PR -- `DocSeeker` fails to precompile on the current nightly: https://travis-ci.org/JunoLab/Atom.jl/jobs/589374224

So I'm going to merge this PR for now.

src/completions.jl
```diff
@@ -101,58 +101,45 @@ returntype(mod, line, c::REPLCompletions.MethodCompletion) = begin
...
-  strlimit(typ, 20)
-end
-returntype(mod, line, c::REPLCompletions.FieldCompletion) = begin
-  typ = string(fieldtype(c.typ, c.field))
-  strlimit(typ, 20)
+  shortstr(inf)
 end
-returntype(mod, line, ::REPLCompletions.PathCompletion) = "Path"
+completionreturntype(c::REPLCompletions.PropertyCompletion) =
+  getproperty(c.value, c.property) |> typeof |> shortstr
```
pfitzseb: :(

aviatesk: 😱

PR #188, opened 2019-09-26
hide macro calls in outline

pfitzseb:


test/display.jl
```diff
@@ -7,7 +7,7 @@
     @test length(Atom.trim(rand(50), 20)) == 20
 
     @test Atom.isanon(sin) == false
-    @test Atom.isanon(x -> x)
+    @test Atom.isanon(x -> x) == true
```
aviatesk: I would really appreciate the comment here saying that the travis would fail otherwise ?

pfitzseb: I don't think the `== true` is relevant, but if tests pass I will add a comment.

PR #192, opened 2019-09-27
enhanced symbols

aviatesk:
### Purposes

1. make `julia-client: goto-symbol` really `goto-symbol`
  * [x] goto local symbols
  * [x] goto toplevel symbols

2. use the same logic in datatip
  * [x] local symbol datatip
  * (toplevel bindings are already coverted by previous functionality)

#### goto

algorithm flow
1. look for the local symbols first for a context block
2. look for the toplevel symbols in a context module, caching searched symbols
3. look for the methods (the previous functionality)

how toplevel symbols search works:
1. find an entry file for a context module (fallback to the active text editor file if `Base.find_package` fails)
2. collect toplevel symbols in a file, and caches them in `symbolscache`
  - so caches are kept _per file_
  - they will get refreshed after the file has been modified
3. search _toplevel_ `include`s, and for each file, recursively call step 2
  - we may want recursive call search, which enables us finding nested calls

#### datatip

flow:
1. look for the local symbols first for a context block
2. show toplevel bindings (already covered by the previous implementation)

### Discussion

`toplevel_bindings` is now renamed into `toplevelitems`, and it finds both toplevel bindings/calls:
- But we may want to know nested calls in some cases, e.g.: `include` & `@testset`
  * which enables finding every `include` calls and describing the `@testset`
- So keeping the previous `toplevel_bindings` and creating a new separate function like `calls(expr, text, call = [], line = 1, pos = 1)` can be appreciated, but I wonder it makes each downstream (outline, goto) slow

Caches are kept _per file_:
- This is simple and robust choice, but this doesn't work when called from "docpane" or "workspace" for modules where `Base.find_package` doesn't work
  * e.g.: When we click a symbol in `Base.Docs` module from its workspace view, because `Base.find_package("Base.Docs")` doesn't work and there is no fallback path, so the goto fails
- If we cache symbols _per module_, this can be improved -- the goto from "workspace" will work after the symbols are cached by a `goto` call with an active editor file
  * With this direction, we may need to handle "module changing" in each file in order to avoid false gotos, but this won't happen so often, and even we don't handle this, it won't be so annoying I guess
  * Other than that, I didn't come up with the pitfalls with this direction, and will try to implement. But atst I want to hear your ideas


src/datatip.jl
```diff
@@ -8,21 +8,59 @@ handle("datatip") do data
...
+  datatip(word, mod, path, column, row, startRow, context)
+end
+
+function datatip(word, mod, path, column = 1, row = 1, startRow = 0, context = "")
   if isdebugging() && (datatip = JunoDebugger.datatip(word, path, row, column)) !== nothing
     return Dict(:error => false, :strings => datatip)
   end
 
+  let localdatatip = localdatatip(word, column, row, startRow, context)
+    localdatatip !== nothing && return Dict(:error => false, :line => localdatatip)
```
aviatesk: So the datatip is going to show the line where `word` is defined. Obviously it's not complete -- the whole binding won't be covered for some cases like
```jl
val = begin
  ...
end
```
(with the current implementation the datatip would only show `val = begin`)

As an alternative, I tried to implement a function that shows the whole binding from `CSTParser.Binding`, but it didn't work well ...

src/goto.jl
```diff
@@ -0,0 +1,84 @@
+handle("goto") do data
+  @destruct [
+    word,
+    path || "",
+    mod || "Main",
+    column || 1,
+    row || 1,
+    startRow || 0,
+    context || "",
+    onlytoplevel || false
```
aviatesk: `onlytoplevel` gonna be `true` when it's called from the doc pane, where there is no local context

src/datatip.jl
aviatesk: addressed in d3b5c5a

pfitzseb:
Awesome, this is great!

> If we cache symbols per module, this can be improved -- the goto from "workspace" will work after the symbols are cached by a goto call with an active editor file
>   - With this direction, we may need to handle "module changing" in each file in order to avoid false gotos, but this won't happen so often, and even we don't handle this, it won't be so annoying I guess
>   - Other than that, I didn't come up with the pitfalls with this direction, and will try to implement. But atst I want to hear your ideas

Yeah, this is definitely the right way to do this. Something like [`Revise.modulefiles`](https://github.com/timholy/Revise.jl/blob/0369acd5ab5fc662749d9b74775216412d33c132/src/pkgs.jl#L84-L101) seems like it'd be useful for this (and in general, I suppose).

[This code](https://github.com/JunoLab/Atom.jl/pull/192/files#diff-02368cc6c68af34a0c260da25e6b4740R108-R111) (kinda) duplicates what CodeTools.jl is trying to do right now, but probably works better. Would probably make sense to use the same logic for module auto-detection etc.

src/outline.jl
```diff
@@ -326,8 +330,12 @@ function str_value(x)
         return string("\"\"\"", x.val, "\"\"\"")
     elseif x.kind === CSTParser.Tokens.STRING
         return string("\"", x.val, "\"")
+    elseif x.kind === CSTParser.Tokens.EQ
+        return " = "
+    elseif x.kind === CSTParser.Tokens.WHERE
+        return " where "
     elseif x.typ === CSTParser.Parameters
-        return ";" * join(str_value(a) for a in x)
+        return "; " * join(str_value(a) for a in x)
```
aviatesk: I made these changes since strings from `str_value` now can be shown in ink.goto.goto items, and it would make more consistent with the strings rendered `Method`s.

But you mayn't like this, in especially in outline.
If so, I will revert this change, so let me know.

pfitzseb: Let's keep it, I've been annoyed by this too. I'm sure there must be a "correct" way to handle whitespace, but this is good enough for me.

pfitzseb: Let's keep it, I've been annoyed by this too. I'm sure there must be a "correct" way to handle whitespace, but this is good enough for me.

pfitzseb:
Btw, I was wondering if it'd make sense to expand the `@testset` functionality (which is great to have) to any toplevel `@macro ...` call. That'd mean that our own
```
@render ::Inline ...
```
definitions would show up, but would also be useful for stuff like
```
@generated f(x) = x
```

src/outline.jl
aviatesk: okay, then let's leave implementing in a "correct" way for the future :p

aviatesk:
It would make sense to me, too.

I'm just wondering whether to accept macro calls without arguments, like
```jl
...
@nospecialize
...
```,
but it would also show _incomplete_ calls as well.

**Comment by pfitzseb** [comment]
I was thinking of just taking the first line verbatim (and maybe limiting it to ~100chars).

**Comment by aviatesk** [comment]
> I was thinking of just taking the first line verbatim (and maybe limiting it to ~100chars).

Ah yeah, that would be fine. Implemented.

**Comment by aviatesk** [comment]
> This code (kinda) duplicates what CodeTools.jl is trying to do right now, but probably works better. Would probably make sense to use the same logic for module auto-detection etc.

Feeling like this PR has became so fat, so maybe putting this to another PR would be better ?
I think using the same logic (i.e. same functions) for auto-module detections, outlines, symbol jumps would be appropriated in terms of consistency and avoiding JIT overheads, though.

**Comment by pfitzseb** [comment]
For sure, was just pointing that out.

**Comment by aviatesk** [comment]
> Something like Revise.modulefiles seems like it'd be useful for this (and in general, I suppose).

So, we now have 2 alternatives for the module file detection, right ?

1. the current approach -- statically parsing files and looks for `include`s
2. mimic the logic of `Revise.modulefiles`

I think if we have `Revise.modulefiles`, then we can do something like
```jl
cache = map(Base.loaded_modules_array()) do m
    m, modulefiles(m)
end
```
and hooking `using/import` calls.

I'm wondering which way we really should take, both in general, and for this symbol jumps ...

pfitzseb:
Well, `Revise.modulefiles` only works for precompiled packages, so we probably need both implementations.

But yeah, in general we should build a `Module -> Files -> Symbols` mapping and update that when needed (e.g. `onDidStopChanging` regenerates the symbols for a particular file, but *only* that file).

aviatesk:
> Well, Revise.modulefiles only works for precompiled packages, so we probably need both implementations.

Okay, so how about this flow ?
1. call `modulefiles(mod)` first and use the returned files if it succeeds
2. if 1. fails, falls back to `include`s searching via static parsing
(and injecting buffer texts for unsaved files for both steps if they are passed from Atom)

aviatesk:
[This commit](a453a4f) fixes the docstring leaks in outline (the bindings with docs haven't been shown in outline with the previous implementations)

src/goto.jl
```diff
@@ -0,0 +1,274 @@
...
+    GotoItem(text, path, line)
+  end
+end
+localgotoitem(word, ::Nothing, column, row, startrow, context) = [] # when `path` is not destructured
+
+### global goto - bundles toplevel gotos & method gotos
+
+# TODO: shows both original & overloaded method defitions: e.g.: Atom.isconst & Base.isconst
+function globalgotoitems(word, mod, text, path)
+  # heuristic: for Base modules, search methods first, since finding/searching their symbols can take some time
```
aviatesk: Maybe we may want to remove this heuristics, i.e.: will help someone edit `Base` files.

src/goto.jl
```diff
@@ -0,0 +1,274 @@
...
+    parsed = CSTParser.parse(text, true)
+    items = toplevelitems(parsed, text)
+    push!(SYMBOLSCACHE[mod], path => items)
+  else
+    # there is no cache
+    SYMBOLSCACHE[mod] = searchtoplevelitems(mod, text, path)
+  end
+end
+
+## TODO: generate toplevel symbols cache for project modules, like `regenerateCache` for docs
```
aviatesk: Any idea ? https://julialang.slack.com/archives/C6A044SQH/p1570026949051100

src/utils.jl
```diff
@@ -163,3 +163,216 @@ cangetdocs(mod::Module, word::Symbol) =
...
+    parentfile = normpath(Base.find_source_file(parentfile))
+    filedata = Base._included_files
+  else
+    use_compiled_modules() || return nothing, nothing   # FIXME: support non-precompiled packages
+    _, filedata = pkg_fileinfo(id)
+  end
+  filedata === nothing && return nothing, nothing
+  included_files = filter(mf -> mf[1] == mod, filedata)
+  return fixpath(parentfile), [fixpath(mf[2]) for mf in included_files]
+end
```
aviatesk: All the adaption is for `modulefiles` and `parentfile`.

The `src_file_key` stuff is removed, since every modifed files will be caught by `regeneratesymbols` handler.
The other changes are minor, basically making every path to be full path

aviatesk:
I've almost done ...
The only things is to fix test cases in `goto.jl`, I hope.

src/goto.jl
pfitzseb: Are you looking for `Base.find_package("Atom")`?

aviatesk: Why do I forget that.
yes, exactly. thanx.

src/goto.jl
```diff
@@ -0,0 +1,274 @@
...
+    text = l[:name]
+    line = startrow + l[:line] - 1
+    GotoItem(text, path, line)
+  end
+end
+localgotoitem(word, ::Nothing, column, row, startrow, context) = [] # when `path` is not destructured
+
+### global goto - bundles toplevel gotos & method gotos
+
+# TODO: shows both original & overloaded method defitions: e.g.: Atom.isconst & Base.isconst
```
aviatesk: I'm thinking of addressing this by appending `methoditem`s whose `file` isn't identical to any of `toplevelitem`

aviatesk: addressed by 951e1e2

aviatesk:
Okay, I think this PR is ready for merge.

PR #193, opened 2019-09-28
don't error on `ModuleCompletion`s for module where `@doc` isn't defined

aviatesk:


pfitzseb:
Isn't this the "wrong" fix and we should be using something like `Main.@doc` [here](https://github.com/JunoLab/Atom.jl/pull/193/files#diff-de08f51fe93b9406046593271d4e89c3L155)?

aviatesk:
Sorry for hasty merge.

Is there any case that there are docs where `@doc` is not defined ?

If so, maybe changing [this line](https://github.com/JunoLab/Atom.jl/blob/master/src/utils.jl#L155) to something like
```jl
word = "$mod.$word"
Core.eval(Main, :(@doc($(Symbol(word)))))
```
will help ?

aviatesk:
> Is there any case that there are docs where @doc is not defined ?

Ah yeah, we can attach docs to objs after they gets defined.

I will make a PR to fix this

pfitzseb:
No, not all modules are loaded in `Main` (e.g. package dependencies -- even though Atom depends on CodeTools, you can't access `Main.CodeTools`).

> Is there any case that there are docs where `@doc` is not defined ?

Pretty sure that e.g. a `baremodule` doesn't have `@doc` defined, but docstrings work just fine.

aviatesk:
> No, not all modules are loaded in Main (e.g. package dependencies -- even though Atom depends on CodeTools, you can't access Main.CodeTools).

Yeah, definitely.

So might be calling `@doc word` from a module [_where completions are invoked_](https://github.com/JunoLab/Atom.jl/blob/master/src/completions.jl#L108-L132) is a possible solution, right ? (of course, only if `@doc` is defined in the module)

aviatesk:
so something like
```jl
completionsummary(mod, c::REPLCompletions.ModuleCompletion) = begin
  m, word = c.parent, c.mod
  cangetdocs(m, word) || return ""
  docs = if isdefined(m, Symbol("@doc"))
    getdocs(m, word)
  else
    word = string(m) * "." * word
    getdocs(mod, word)
  end
  description(docs)
end
```

PR #195, opened 2019-10-07
improve error printing

pfitzseb:
and add an option to always show errors in REPL when using inline evaluation.

Fixes https://github.com/JunoLab/Juno.jl/issues/381.

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -52,13 +52,12 @@ handle("evalshow") do data
     result = hideprompt() do
       with_logger(JunoProgressLogger()) do
         withpath(path) do
-          try
-            res = include_string(mod, text, path, line)
-            res ≠ nothing && !ends_with_semicolon(text) && display(res)
-            res
-          catch e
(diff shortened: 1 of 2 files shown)


src/eval.jl
```diff
@@ -117,21 +124,17 @@ handle("evalall") do data
...
-              println(stderr)
-            else
-              render(Console(), ee)
-            end
+          catch err
+            ee = EvalError(err, st)
+
+            # show error in REPL:
+            Base.showerror(IOContext(stderr, :limit => true), ee)
+            # show notification (if enabled in Atom):
```
aviatesk: This is also the previous behaviour, but why we always don't show errors inline with `evalall` ?

pfitzseb: `evalall` shouldn't show anything inline, imho -- it's basically a shorthand for `include(currentFile)`.

aviatesk: Okay, understand.
Other than that everything looks good to me.

PR #196, opened 2019-10-10
prevent items inside quotes from leaking into outline

aviatesk:


test/outline.jl
```diff
@@ -286,6 +286,20 @@ end
         end
     end
 
+    # the items inside quotes doesn't leak
```
pfitzseb:
```suggestion
    # items inside quote blocks don't leak
```

PR #199, opened 2019-10-21
re-use `basepath` for getting path to julia install

pfitzseb:
`isdir` can error when the user doesn't have read access to `basebuilddir`, and afaict the code to build that const isn't really necessary anyways (or at least noone has complained about our `basepath` implementation).

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -219,7 +219,7 @@ end
 function fixpath(
   filename::AbstractString;
   badpath = basebuilddir,
-  goodpath = juliadir,
+  goodpath = basepath("..")
 )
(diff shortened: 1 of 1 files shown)


aviatesk:
thanks for the fix.

and then we no longer need:
https://github.com/JunoLab/Atom.jl/blob/52122c87ce439c203f3183a6f681bb428c1fdaa1/src/utils.jl#L243-L252

pfitzseb:
No, we still need that to find the "bad path" that `fixpath` fixes.

aviatesk:
ah my bad, I missed it.

PR #200, opened 2019-10-22
fixes JunoLab/Juno.jl#404

aviatesk:
A `where` clause creates `LocalScope`, which holds method arguments as its children, but `_actual_local_bindings` fails to enter into its children on https://github.com/JunoLab/Atom.jl/blob/1f9cf9db6020a84a5b7430850d2cd8c9bf1cd35c/src/outline.jl#L303

Fixes via the same approach as https://github.com/JunoLab/Atom.jl/pull/183


aviatesk:
the travis failure seems to be identical to https://github.com/JuliaLang/julia/issues/30555

pfitzseb:
Restarted the build.

PR #202, opened 2019-10-23
rename refactor

aviatesk:
### Purpose

Implement rename refactoring command:
- rename local binding
- rename global bindings (module-based)

### Approach

- use MacroTools.jl for easy tokenizing and code replacing
- for module-based global refactoring, CSTParser-based static file detection is used:
  * since our Revise-like module file detection will easily fail with the refactorings, which breaks precompilation cache
- first tries local refactoring, and then move to global refactoring if needed
- **rename refactoring on fields is all suppressed, since not every code is statically typed and so I think we would end up with lots of false positive**
- global refactoring are designed to work only when it's called on the definition place

### Task / Achievements

- [x] local refactor
- [x] module-based global refactor
- graceful messages
  * [x] notice the need of refactoring in the parent module if the refactored binding is imported/overloaded
  * [x] report back error message
  * [x] notice refactored files
  * [x] show context
- handle edge cases
  * [x] suppress keyword refactoring
  * [x] suppress field refactoring
  * [x] suppress global refactoring that are not called at a definition place
  * [x] halt global refactoring if files without write permission detected
  * [x] global refactoring with dot accessor like `Atom.isdebugging()`, `Base.countlines(args...)`
  * [x] global refactoring on macro
- [ ] test


aviatesk:
xref: https://github.com/JunoLab/atom-julia-client/pull/641
xref: https://github.com/JunoLab/atom-ink/pull/236

aviatesk:
woa, travis has been revived ! :)

src/refactor.jl
```diff
@@ -0,0 +1,230 @@
...
+      end
+
+      desc = join(("_Global_ rename refactoring `$mod.$old` ⟹ `$mod.$new` succeeded.", moddesc, desc), "\n\n")
+    end
+
+    return Dict(kind => desc)
+  catch err
+    return Dict(:error => errdescription(old, new, err))
+  end
+end
```
aviatesk: I agree the logic here would look messy, so please feel free to ask me if you have anything unsure -- I probably miss some edge cases.

aviatesk:
Here are some edge cases I've noticed but still I'm not sure about how to handle.

### local refactoring

If we have the code like:

```julia
function outer()
  val = 10
  function inner(val)
    2 * val
  end
  return inner(val)
end
```

and we refactor on `outer`'s `val`, the `inner`'s `val`s are also refactored, even though we may want to only refactor `outer`s.
This is obviously because the `inner` function is included in the `outer` function scope, but I still not found any good way to handle this case with MacroTools's source walk.

### global refactoring

If a module contains code like:

```julia
function fun() end

function other()
  fun = 10
end
```

and we rename `fun` to `fun2`, then the binding `fun` in `other` will also be renamed to `fun2`. This is because I intentionally looks just for `fun` tokens, not for call sites ([the code here](https://github.com/JunoLab/Atom.jl/blob/b8802f34ded21f937f45be7ee384b70a6abc36a8/src/refactor.jl#L155-L163)), since in Julia every bindings is an object, i.e.: we can have a code like `map(fun, itr)`.
Current behavior can be annoying, but at least doesn't break code. If we just looks for call sites, then we may have false negatives, which would _break_ the code. So imho the current behavior may suffice.

aviatesk:
Damn, I missed to push to the master ! Will try to revert all the commits.

aviatesk:
Okay, sorry for the confusing.
283d46602e024b92fc659d67810c4927d84d713e (on master) successfully reverted all the commits in this PR that were mistakenly pushed to master.
And this PR is moved into https://github.com/JunoLab/Atom.jl/pull/203.

PR #203, opened 2019-10-23
WIP: rename refactor

aviatesk:
### Purpose

Implement rename refactoring command:
- rename local binding
- rename global bindings (module-based)

### Approach

- use MacroTools.jl for easy tokenizing and code replacing
- for module-based global refactoring, CSTParser-based static file detection is used:
  * since our Revise-like module file detection will easily fail with the refactorings, which breaks precompilation cache
- global refactoring are designed to work only when it's called on the definition place
- first tries local refactoring, and then move to global refactoring if needed
- **rename refactoring on fields is all suppressed, since not every code is statically typed and so I think we would end up with lots of false positive**

### Task / Achievements

- [x] local refactor
- [x] module-based global refactor
- graceful messages
  * [x] notice the need of refactoring in the parent module if the refactored binding is imported/overloaded
  * [x] report back error message
  * [x] notice refactored files
  * [x] show context
- handle edge cases
  * [x] suppress keyword refactoring
  * [x] suppress field refactoring
  * [x] handle global refactoring that are not called at a definition place
  * [x] handle global refactoring on an unsaved editor
  * [x] handle global refactoring on files without write permission detected
  * [x] global refactoring with dot accessor like `Atom.isdebugging()`, `Base.countlines(args...)`
  * [x] global refactoring on macro
- [ ] test


aviatesk:
xref: https://github.com/JunoLab/atom-julia-client/pull/641
xref: https://github.com/JunoLab/atom-ink/pull/236

aviatesk:
Here are some edge cases I've noticed but still I'm not sure about how to handle.

### local refactoring

If we have the code like:

```julia
function outer()
  val = 10
  function inner(val)
    2 * val
  end
  return inner(val)
end
```

and we refactor on `outer`'s `val`, the `inner`'s `val`s are also refactored, even though we may want to only refactor `outer`s.
This is obviously because the `inner` function is included in the `outer` function scope, but I still not found any good way to handle this case with MacroTools's source walk.
As an alternative, this case would be easily solved if we can directly manipulate CSTParser's expressions and then reconstruct the source _appropriately_. But I'm not sure even if it's possible.

### global refactoring

If a module contains code like:

```julia
function fun() end

function other()
  fun = 10
end
```

and we rename `fun` to `fun2`, then the binding `fun` in `other` will also be renamed to `fun2`. This is because I intentionally looks just for `fun` tokens, not for something more complicated like call sites ([the code here](https://github.com/JunoLab/Atom.jl/blob/b8802f34ded21f937f45be7ee384b70a6abc36a8/src/refactor.jl#L155-L163)), since in Julia every binding is an object, and we can have a code like `map(fun, itr)`.
Current behavior can be annoying, but at least doesn't break code. If we just looks for call sites, then we may have false negatives, which would _break_ the code. So imho the current behavior may suffice.

aviatesk:
I agree the logic inside [`renamerefactor`](https://github.com/JunoLab/Atom.jl/blob/86b3b2c795cda7a3accd99b7b72476e3bf36040a/src/refactor.jl#L18-L77) would look messy, and please feel free to ask me if you find anything weird -- I probably miss some edge cases.

https://github.com/JunoLab/Atom.jl/blob/86b3b2c795cda7a3accd99b7b72476e3bf36040a/src/refactor.jl#L18-L77

aviatesk:
#208, #209, #210 need to be merged in order to finish up this PR (this only thing remain is to add tests)

pfitzseb:
Both of these edge cases should be avoidable by checking each new local scope for assignments that shadow a global, right? We already should have the logic for that.

pfitzseb:
We'll also need to think about proper UIs for all of this, especially the global refactors. I'd love to reuse `find-and-replace`s pane, but that doesn't seem to be possible.

So we could/should build one ourselves and ideally make it reuseable for something like [this](https://github.com/facebookarchive/atom-ide-ui/blob/master/docs/find-references.md).

aviatesk:
> We'll also need to think about proper UIs for all of this, especially the global refactors. I'd love to reuse `find-and-replace`s pane, but that doesn't seem to be possible.
> 
> So we could/should build one ourselves and ideally make it reuseable for something like [this](https://github.com/facebookarchive/atom-ide-ui/blob/master/docs/find-references.md).

Yeah, would be great to have, and I'm willing to implement that.
To me it seems better to start with implementing the "Find reference" feature, and then enable the global rename-refactor within that as an additional feature.

aviatesk (commented):
Needs https://github.com/MikeInnes/SourceWalk.jl/pull/1 to be merged for this to work.

PR #208, opened 2019-10-29
strip trailing dots in goto/datatip (and rename)

aviatesk:
While implementing #203, I had to implement the logic to remove trailing dots and found that this is better to be merged as a separate feature, since the logic can be used for goto/datatip.

In short, if we have a word like
```jl
Atom.SYMBOLSCACHE.age
```
and then, we can get goto and datatip for:
- `Atom` when our cursor is anywhere on `Atom`
- `Atom.SYMBOLSCACHE` when our cursor is anywhere on `SYMBOLSCACHE`
- no datatip if our cursor is on `age` (the field)

(c.f.: The current behaviour: we can't get any goto and datatip at all)

***

Requires https://github.com/JunoLab/atom-julia-client/pull/651

### DIFF
--- a/src/datatip.jl
+++ b/src/datatip.jl
@@ -6,7 +6,6 @@ atom-ide-ui is already deprecated, ugly, not fully functional, and and...
 handle("datatip") do data
   @destruct [
     word,
-    fullWord,
     mod || "Main",
     path || "",
     column || 1,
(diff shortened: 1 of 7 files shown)


aviatesk:
@pfitzseb any objection on this idea ? I will make changes so that the striping would be done in the frontend, though.

aviatesk:
okay, so the striping task is now all done in the frontend ([basically by using different regex for the beginning and end of the word](https://github.com/JunoLab/atom-julia-client/blob/2edaf2b2de5cc18d25f9b2f166c2de4d60ac03d4/lib/misc/words.js#L54-L103)), and the awkward `fullword` stuff here are all removed.

>  Might be somewhat unintuitive

yes, maybe especially when we get goto/datatip for `Mod` when a cursor is on `Mod.foo` or something.
(But at least there are corresponding underline/highlights)

***

Still I'm not sure which behaviour do people prefer (not so many of users mayn't even care about the difference tho)

aviatesk:
The last two commits refactors the logic of `gotoglobalsymbols` and updates tests according to that.

Once the CI passes, I will merge this PR, since now stripping trailing dots is all done in the frontend (JunoLab/atom-julia-client#651) and the changes in this PR only do trivial refactors.

PR #210, opened 2019-10-29
setup modules.jl

aviatesk:
- bundles module-detection code into modules.jl
  * finding module files
  * auto-module detection
- add some basic tests for them

***

- Separated from #203 for easier review
- Also prepares for https://github.com/JunoLab/Juno.jl/issues/411


aviatesk:
#209 needs to be merged in advance.

PR #215, opened 2019-11-03
RFC: module validation check within CSTParser-based module traverse

aviatesk:
### main purpose

fixes module escaping issues within CSTPraser-based module traverse
- [x] fixes that it can enter into submodules when searched from the parent module
- [x] fixes that it can include items outside of modules in the entry file (e.g. [`isdebugging` in `Atom` module](https://github.com/JunoLab/Atom.jl/blob/master/src/debugger/debugger.jl#L13) when we traverse `Atom.JunoDebugger`)

That would make goto and rename refactors (https://github.com/JunoLab/Atom.jl/pull/203) more "correct", and may also help https://github.com/JunoLab/Juno.jl/issues/411.

### sub note

I also included changes that enables gotos within unsaved editors

### DIFF
--- a/src/goto.jl
+++ b/src/goto.jl
@@ -194,8 +194,7 @@ function _collecttoplevelitems!(mod::Union{Nothing, String}, entrypath::String, 
   _collecttoplevelitems!(mod, entrypath, text, pathitemsmaps)
 end
 function _collecttoplevelitems!(mod::Union{Nothing, String}, entrypath::String, text::String, pathitemsmaps::PathItemsMaps)
-  parsed = CSTParser.parse(text, true)
-  items = toplevelitems(parsed, text; mod = mod)
+  items = toplevelitems(text; mod = mod)
   push!(pathitemsmaps, entrypath => items)
 
   # looking for toplevel `include` calls
--- a/src/goto.jl
+++ b/src/goto.jl
@@ -1,5 +1,3 @@
-using CSTParser
-
 handle("gotosymbol") do data
   @destruct [
     word,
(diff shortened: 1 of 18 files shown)


aviatesk:
Updated the description.

aviatesk:
> [ ] fixes that it can include items outside of modules in the entry file (e.g. isdebugging in Atom module when we traverse Atom.JunoDebugger)

As for this issue, I first tried to fix this by introducing a flag for `toplevelitems` that indicates whether it's in a target module (and update that appropriately in each downstream), but I found that would increase code complexity (somewhat much) more -- and I think creating a struct `ToplevelScope` like `LocalScope` would also end up introducing the same kind of complexity. 

I may try to implement that anyway, but also want to hear better ideas if there are.

aviatesk:
> > [ ] fixes that it can include items outside of modules in the entry file (e.g. isdebugging in Atom module when we traverse Atom.JunoDebugger)
> 
> I first tried to fix this by introducing a flag argument for `toplevelitems` that indicates whether it's in a target module (and update that appropriately in each downstream), but I found that would increase code complexity somewhat much more -- and I think creating a struct `ToplevelScope` like `LocalScope` would also end up introducing the same kind of complexity.
> 
> I may try to implement that anyway, but also want to hear better ideas if there are.

Okay, I've implemented the fix along with this way within the latest commit.
The code comes to look more complex, but at least imo functions APIs are _acceptable_.

Now I've fixed all the issues that I'm trying to address within this PR, and I would welcome review or comment :)

src/goto.jl
```diff
@@ -222,34 +240,43 @@ function GotoItem(path::String, bind::ToplevelBinding)
...
+
+  # initialize the cache if there is no previous one
+  if !haskey(SYMBOLSCACHE, mod)
+    SYMBOLSCACHE[mod] = collecttoplevelitems(m, path, text)
   end
+
+  # ignore toplevel items outside of `mod` when `path` is an entry file
+  entrypath, _ = moduledefinition(m)
+  inmod = path != entrypath
+  items = toplevelitems(text; mod = stripdotprefixes(mod), inmod = inmod)
```
aviatesk: Now there are at most 2 calls of `toplevelitems` in `updateeditor` handler, since we don't want module escaping in outline but do want in symbols update.

pfitzseb:
Looks good, nice work! 
There will of course still be cases this doesn't handle properly (like imported bindings), but that's why we have the runtime stuff as a fallback.

aviatesk:
@pfitzseb 
Do you have any alternative idea instead of the new function API with those keyword arguments ? I'm still feeling like they're rather complicated, so I really want to hear if you have the other idea.

pfitzseb:
No, but I'll let you know if I figure out something better ;)

PR #217, opened 2019-11-11
follow https://github.com/timholy/Revise.jl/pull/385

aviatesk:
Later I will open an issue on Revise's repo about the discussion on extracting upstream code, which would get rid of the need for these changes.

### DIFF
--- a/src/modules.jl
+++ b/src/modules.jl
@@ -116,17 +116,32 @@ function pkg_fileinfo(id::PkgId)
   # Try to find the matching cache file
   paths = Base.find_all_in_cache_path(id)
   sourcepath = Base.locate_package(id)
-  for path in paths
-    Base.stale_cachefile(sourcepath, path) === true && continue
-    provides, includes_requires = parse_cache_header(path)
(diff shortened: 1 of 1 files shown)


pfitzseb:
We should also check if we can't make use of CodeTracking.jl more (that can provide files included in a loaded package, IIRC).

aviatesk:
yeah, we should. I will check the code base before opening an issue today.

aviatesk:
I took a quick look at CodeTracking.jl's code base (and Revise's), and now I guess there is no strong motivation for us to use the CodeTracking's file-detection logic:
- first, we rather want to files in "module"s, not ones in a package, most of the times
- actually Revise seems to only use the logic of CodeTracking to _store_ package files information, and update that by finding the files by Revise itself (and the CodeTracking's logic to _find_ files in a loaded package like `CodeTracking.pkgfiles` is no longer used, and it even seems to be kinda stale ?)

Anyway, I think we're ready to discuss this and let's file an issue.

PR #219, opened 2019-11-22
more memory efficient SYMBOLSCACHE

aviatesk:
Don't let `SYMBOLSCACHE` eat memory like a horse -- now it only uses`4 MiB` after `julia-client: regenerate-symbol-cache` while it did `140 MiB` within the previous implementation.

refactors:
- only keep _simple_ data in `SYMBOLSCACHE`
- remove `ToplevelTupleH` struct: make the overall code much simpler

### DIFF
--- a/src/goto.jl
+++ b/src/goto.jl
@@ -220,15 +220,6 @@ function filtertoplevelitem(word, bind::ToplevelBinding)
   bind = bind.bind
   bind === nothing ? false : word == bind.name
 end
-function filtertoplevelitem(word, tupleh::ToplevelTupleH)
-  expr = tupleh.expr
-  for arg in expr.args
-    if str_value(arg) == word
-      return true
-    end
-  end
-  return false
-end
 
 function GotoItem(path::String, bind::ToplevelBinding)
   expr = bind.expr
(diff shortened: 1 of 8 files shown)


src/goto.jl
```diff
@@ -109,41 +113,54 @@ end
...
-  for (path, items) in pathitemsmaps
-    for item in filter(item -> filtertoplevelitem(word, item), items)
-      push!(ret, GotoItem(path, item))
-    end
+  for (_, items) in pathitemsmap
+    @>> filter(items) do item
+      let item = item
+        word == item.name
+      end
+    end append!(ret)
```
aviatesk: sorry for this `@>>` macro, but I couldn't come up with the appropriate name for this

src/goto.jl
```diff
@@ -109,41 +113,54 @@ end
...
+
+# NOTE:
+# this data structure should only keep "simple" data so that it doesn't eat
+# too much memory when the cache has got fat (like via `regeneratesymbols`)
+# TODO?:
+# maybe we will end up creating a more generic struct that represents a "symbol",
+# and wanting `SYMBOLSCACHE` to store them instead of `GotoItem`
+# (should be kept "simple" atst though, and ideally it handles both toplevel and local symbols)
+# so that we can use `SYMBOLSCACHE` as a general cache interface for various static features
+# like goto, find references, and so on.
```
aviatesk: I believe this would be the better and reusable implementation, but let's leave this as a future ref in this PR.

PR #220, opened 2019-11-22
outline module usages

aviatesk:
create `ToplevelModuleUsage` struct and shows them in outline:
image

we may want to use them in modules.jl or goto.jl as well for more precise module traverse in the future

***

First I would like to merge #219 and the rebase this PR.

### DIFF
--- a/src/static/toplevel.jl
+++ b/src/static/toplevel.jl
@@ -16,7 +16,13 @@ end
 struct ToplevelCall <: ToplevelItem
     expr::CSTParser.EXPR
     lines::UnitRange{Int}
-    callstr::String
+    str::String
+end
+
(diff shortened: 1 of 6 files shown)


src/static/toplevel.jl
```diff
@@ -19,7 +19,7 @@ struct ToplevelCall <: ToplevelItem
     callstr::String
 end
 
-struct ToplevelTupleH <: ToplevelItem
+struct ToplevelModuleUsage <: ToplevelItem
```
aviatesk: please let me know if you have suggestions for this naming

pfitzseb: This is fine, imho.

PR #222, opened 2019-11-22
fix CI failure on v1.4

aviatesk:


aviatesk:
ref: https://github.com/JuliaLang/julia/pull/33477

Issue #226, opened 2019-12-01
update to CSTParser@1.1.0


aviatesk:
as discussed [here](https://github.com/JuliaRegistries/General/pull/6019), we should recover binding/scope stuff that are remove in CSTParser@1.1.0 by ourselves somehow.

Will add upper bound versioning later for now.

PR #227, opened 2019-12-01
temporal remedy for #226

aviatesk:


aviatesk:
merged this so that the current other PRs can use CI for now.
Should be reverted to the usual semver specification once #226 is fixed

PR #229, opened 2019-12-10
use singleton instance `undefined`

aviatesk:


aviatesk:
- fix `getfield′`
- add tests for `getfield′`
- add test for `undefined`, `isundefind`

PR #230, opened 2019-12-12
misc

aviatesk:


src/completions.jl
```diff
@@ -165,7 +165,6 @@ completiontype(c) = "variable"
 completiontype(c::REPLCompletions.ModuleCompletion) = begin
   ct = completion_text(c)
   ismacro(ct) && return "snippet"
-  ct == "Vararg" && return ""
```
aviatesk: Not sure why this special casing was introduced, but I believe `Vararg`'s completion type is better to be expressed as `type`.

May try to find the commit where the rationale was explained.

aviatesk: okay, original this was introduced by @pfitzseb within https://github.com/JunoLab/Atom.jl/pull/125/commits/3be3770c07125b0774582304237e967c1480a592.

Could you please explain me the reason for that ?

pfitzseb: Probably because of
```
julia> REPL.REPLCompletions.get_type(:Vararg, Main)
ERROR: TypeError: in Type, in parameter, expected Type, got Vararg
Stacktrace:
 [1] Typeof at ./boot.jl:216 [inlined]
 [2] typesof(::Type) at ./errorshow.jl:166
 [3] get_type(::Symbol, ::Module) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/REPL/src/REPLCompletions.jl:427
 [4] top-level scope at REPL[5]:1

julia> REPL.REPLCompletions.get_type(:Int, Main)
(Type{Int64}, true)
```

aviatesk: okay, I guess why `Type{Vararg}` isn't a valid julia expression is maybe that `Vararg` should only be used within `Tuple` type and it's not a valid type by itself, right ?

Still I think it's better to be shown as "type" in Juno because it's "special" but surely used for specifying types (with wrapped in `Tuple`).
But atst I want to hear your opinion on this.

pfitzseb: Yeah, it's fine to show this the same as other types.

PR #232, opened 2019-12-18
RFC: better documentation search

aviatesk:
The main change is to improve documentation searches where the needle is like `Mod.needle`.
If `Mod` is available, this would automatically search `needle` in `Mod` module (the same logic as https://github.com/JunoLab/Atom.jl/pull/208)

This would overwrites an user's module specification (if specified),
(e.g.:
image
would overwrite `Atom.JunoDebugger` spec to `JuliaInterpreter`)
but I believe that would be more preferable for reasons:
- an user intentionally type module prefix for these cases
- some documentation references includes module prefixes and they don't work as expected with the previous behaviour:
  * e.g: https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/master/src/construct.jl#L353

***

I also added some tests for this and other.

### DIFF
--- a/src/docs.jl
+++ b/src/docs.jl
@@ -13,7 +13,7 @@ handle("searchdocs") do data
     needle;
     # kwargs to be passed to `DocSeeker.searchdocs`:
     # TODO: configuration for `maxreturns`
-    loaded = !allPackages, mod = mod, exportedonly = exportedonly, name_only = name_only
+    loaded = !allPackages, mod = getmodule(mod), exportedonly = exportedonly, name_only = name_only
   )
 end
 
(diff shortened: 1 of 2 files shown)


pfitzseb:
Hm, as is I would expect this to search for docs in `Juno.JunoDebugger.JuliaInterperter`. What do you thing about automatically deleting the module input if we find a module in the main search field?

aviatesk:
> What do you thing about automatically deleting the module input if we find a module in the main search field?

That would be okay, as far as we _delete_ it (implicit modification (, which would remain) could be annoying I guess).

> I would expect this to search for docs in `Juno.JunoDebugger.JuliaInterperter`

I think we can kinda implement this behaviour by modifying [DocSeeker's filtering logic a bit](https://github.com/pfitzseb/DocSeeker.jl/blob/master/src/introspective.jl#L25-L38). Ideally everything would be straightforward if [`modulebindings`](https://github.com/pfitzseb/DocSeeker.jl/blob/master/src/introspective.jl#L43-L64) handles `topmod` correctly (i.e. correctly finds all the bindings that can be reached from `topmod`), but it seems not for now.
Instead we can pass `topmod` to `dynamicsearch`, and filters results if `mod` isn't defined in `topmod`.
What do you think about this idea ? Maybe showing code would be easier to understand this though.

aviatesk:
d5e7555 and the corresponding https://github.com/pfitzseb/DocSeeker.jl/pull/17 would show my idea:
Now module filtering is more _sensible_ and shows bindings in the other module if they can be reached from the specified module.

e.g.:
image

aviatesk:
> What do you thing about automatically deleting the module input if we find a module in the main search field?

I implemented this locally, and I feel this is okay.

aviatesk:
@pfitzseb I think now you can review on this again with https://github.com/JunoLab/DocSeeker.jl/pull/17 and https://github.com/JunoLab/atom-julia-client/pull/666 -- I believe the module-specified search improves by them.

pfitzseb:
I think I like this, so let's get it into the release.

aviatesk:
hehe, I've forgot these PRs. Thanks for the merges.

PR #234, opened 2019-12-20
RFC: improve completion performance

aviatesk:
needs https://github.com/JunoLab/atom-julia-client/pull/663

***

### Rationale:

A profiling on `basecompletionadapter` shows [`completionsummary`](https://github.com/JunoLab/Atom.jl/blob/master/src/completions.jl#L108-L126) took much time -- this is obvious because retrieving a doc string is done by calling [`include_string` in `getdocs`](https://github.com/JunoLab/Atom.jl/blob/avi/optimcomps/src/utils.jl#L217-L234).
As a consequence, sometimes I can't get completions because of [the racing in frontend](https://github.com/JunoLab/atom-julia-client/blob/avi/optimcomps/lib/runtime/completions.js#L50) (which is good thing though) when there are many of suggestion (say, >100).

So I would like to simply suppress them when there are so many of completion suggestions, because I guess showing completion descriptions when there are many of them is not so usuful.
(as an heuristic, I made an threshold to [30](https://github.com/JunoLab/Atom.jl/blob/avi/optimcomps/src/completions.jl#L78) but this is obviously subject to discussion)

And autocomplete-plus only shows [at most 200 suggestions](https://github.com/atom/autocomplete-plus/blob/master/lib/suggestion-list-element.js#L49), so I made a corresponding threshold in the backend as well so that we can avoid useless computation.

### Achievement

On my machine `basecompletionadapter` now runs 6x times better (and defeating in the race never happens after JIT compiles)

### Misc

As a general improvement, I did some rename refactors (that would make namings more consistent with the other part of code bases), and changed the order of pushing local completions (since `pushfirst` isn't optimal)

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -14,7 +14,7 @@ handle("completions") do data
   ] = data
 
   withpath(path) do
-    comps, prefix = basecompletionadapter(
+    basecompletionadapter(
       # general
       line, mod,
       # local context
@@ -22,8 +22,6 @@ handle("completions") do data
       # configurations
       force
     )
-
-    Dict(:completions => comps, :prefix => prefix)
   end
 end
 
@@ -69,7 +67,7 @@ function basecompletionadapter(
   comps = if force || !isempty(prefix)
     filter!(let p = prefix
       c -> startswith(c[:text], p)
-    end, localcompletions(context, row, column))
+    end, localcompletions(context, row, column, prefix))
   else
     CompletionSuggetion[]
   end
@@ -81,23 +79,24 @@ function basecompletionadapter(
       c isa REPLCompletions.PackageCompletion || continue
     end
     try
-      push!(comps, completion(mod, c, suppresss))
+      push!(comps, completion(mod, c, suppresss, prefix))
     catch err
       continue
     end
   end
 
-  return comps, prefix
+  return comps
 end
 
-completion(mod, c, suppress) = CompletionSuggetion(
+completion(mod, c, suppress, prefix) = CompletionSuggetion(
   :type               => completiontype(c),
   :icon               => completionicon(c),
   :rightLabel         => completionmodule(mod, c),
   :leftLabel          => completionreturntype(c),
   :text               => completiontext(c),
   :description        => completionsummary(mod, c, suppress),
-  :descriptionMoreURL => completionurl(c)
+  :descriptionMoreURL => completionurl(c),
+  :replacementPrefix  => prefix
 )
 
 completiontext(c) = completion_text(c)
@@ -219,12 +218,13 @@ end
 completionicon(::REPLCompletions.DictCompletion) = "icon-key"
 completionicon(::REPLCompletions.PathCompletion) = "icon-file"
 
-localcompletions(text, row, col) = localcompletion.(locals(text, row, col))
-localcompletion(l) = CompletionSuggetion(
-  :type        => l[:type] == "variable" ? "attribute" : l[:type],
-  :icon        => l[:icon] == "v" ? "icon-chevron-right" : l[:icon],
-  :rightLabel  => l[:root],
-  :leftLabel   => "",
-  :text        => l[:name],
-  :description => ""
+localcompletions(text, row, col, prefix) = localcompletion.(locals(text, row, col), prefix)
+localcompletion(l, prefix) = CompletionSuggetion(
+  :type              => l[:type] == "variable" ? "attribute" : l[:type],
+  :icon              => l[:icon] == "v" ? "icon-chevron-right" : l[:icon],
+  :rightLabel        => l[:root],
+  :leftLabel         => "",
+  :text              => l[:name],
+  :description       => "",
+  :replacementPrefix => prefix
 )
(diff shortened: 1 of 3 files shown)


src/completions.jl
```diff
@@ -1,73 +1,107 @@
...
+    # local context
+    context || "",
+    row || 1,
+    startRow || 0,
+    column || 1,
+    # configurations
+    force || false
+  ] = data
 
   withpath(path) do
```
aviatesk: Not sure why do we need `withpath` here. Do you think we can remove this ?

pfitzseb: I *think* I wanted path completions to be relative to the file you're in, but `withpath` doesn't help with that afaict. So yeah, we can probably remove this.

pfitzseb:
> So I would like to simply suppress them when there are so many of completion suggestions, because I guess showing completion descriptions when there are many of them is not so usuful.

Ideally we'd provide baseline suggestions and then use autocomplete-plus's `getSuggestionDetailsOnSelect` to load the docstring and inferred return type.

aviatesk:
> Ideally we'd provide baseline suggestions and then use autocomplete-plus's `getSuggestionDetailsOnSelect` to load the docstring and inferred return type.

ah yeah, definitely. Will give a try to implement that.

src/completions.jl
aviatesk: I get it. Umm maybe we need to modify `withpath` itself then ? If you want to leave this as is, I'm okay with that (relative path completions would be so great)

aviatesk:
> > So I would like to simply suppress them when there are so many of completion suggestions, because I guess showing completion descriptions when there are many of them is not so usuful.
> 
> Ideally we'd provide baseline suggestions and then use autocomplete-plus's `getSuggestionDetailsOnSelect` to load the docstring and inferred return type.

Done the prototyping :)
Will add a comment to parts where I want your help later.

But I think you can test how `getSuggestionDetailsOnSelect` feels like for now.

src/completions.jl
```diff
@@ -187,25 +165,100 @@ end
...
-  # m = isa(m, Module) ? m : Main
-  # CodeTools.completions(m)
+using JuliaInterpreter: sparam_syms
+using Base.Docs
+
+function completiondetail_method!(comp)
+  mod = getmodule(comp[:rightLabel])
+  # more robust way to do this ?
+  f = getfield′(mod, Symbol(comp[:replacementPrefix]))
+  isundefined(f) && return
```
aviatesk: This will be more complex but maybe worth to create an additional key in baseline completions to keep a method's name from `Method.name`, since `prefix` is made in `REPL` module and may change in the future.

src/completions.jl
```diff
@@ -1,74 +1,102 @@
...
-  d = []
-  for c in comps
-    if REPLCompletions.afterusing(line, Int(first(replace))) # need `Int` for correct dispatch on x86
-      c isa REPLCompletions.PackageCompletion || continue
-    end
-    try
-      push!(d, completion(mod, c))
-    catch err
-      continue
-    end
```
aviatesk: I removed this exception check since now our baseline completions are (hopefully) error safe.

aviatesk:
imho I'm still not sure I like the new behaviour, especially about "return type appearing on selection".
After we have decided to go which way to go, I will update test cases.

src/completions.jl
aviatesk: nvm, addressed

aviatesk:
@pfitzseb could you have a look/try at this again ? I really want to hear what do you feel on this change.
This works fairly well, and I also updated tests.

pfitzseb:
I've used this for a couple of days now and it feels pretty good to me. Didn't have too many issues with the previous implementation being too slow though ;)
Will look through the code and then we can merge, I think.

aviatesk:
Thanks for testing by yourself, @pfitzseb :)

> Didn't have too many issues with the previous implementation being too slow though ;)

yeah, I maybe being too picky on this.
(still I'm a little proud of that now a first invocation of method completions with many method instances like `show`, etc won't be slow, for example)

I think this change will complicate things a bit more, so I'm really okay if we're slot to merge this into master.

pfitzseb:
Did you by chance test how expensive getting the return type for a method is? If it's not too bad we could consider always displaying them.

aviatesk:
It really depends on what method you try to complete and whether it's a first time to infer the return type.

For example, if we profile `basecompletionadapter("show(", Main, false, 1, 1, "")` _for the first time_, almost 100% of time is spent on `completionreturntype(c::REPLCompletions.MethodCompletion)` (especially in `typeinf_type`), even if we already compiled `basecompletionadapter` itself.
But after that, the result of the same invocation of `basecompletionadapter("show(", Main, false, 1, 1, "")` will change so much -- now the most of the time (about 70 %) will be spent in `completionsummary`, since the cache of the previous inference will be used for return type inference.

pfitzseb:
Ok, then let's merge this as is.

aviatesk:
let's keep stay flexible to change on this :)

PR #239, opened 2020-01-05
log `cd` handler

aviatesk:
Now `work-in-hoge` commands would be more explicit like `activate-hoge`s :)

image


pfitzseb:
Maybe `Changed working directory to ...` instead? `Working in` seems kinda ambiguous.

aviatesk:
Your suggestion sounds better :) Addressed.

PR #241, opened 2020-01-14
recursive Project.toml search

aviatesk:
Enables `julia-client: activate-environment-in-current-folder` command to work when invoked from any dir under the `Project.toml` file

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -19,9 +19,26 @@ handle("cd") do path
   end
 end
 
-handle("activateProject") do path
+handle("activateProject") do dir
   hideprompt() do
+    if (path = find_project_file(dir)) === nothing
+      @warn "No project file found for `$dir`"
+      return
+    end
     Pkg.activate(path)
+  end
+end
+
+function find_project_file(dir)
+  while true
+    next_dir = dirname(dir)
+    (
+      next_dir == dir || # ensure to escape infinite recursion
+      isempty(dir)       # reached to the system root
+    ) && return nothing
+    path = joinpath(dir, "Project.toml")
+    isfile(path) && return path
+    dir = next_dir
   end
 end
 
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -21,6 +21,12 @@ end
 
 handle("activateProject") do dir
   hideprompt() do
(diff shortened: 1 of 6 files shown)


pfitzseb:
What about when I actually want to activate a directory without a Project.toml?

aviatesk:
Shouldn't work. I've not checked `Pkg.activate(path)` works for directories that only contain `REQUIRE`, but if that works, we can check one more return condition check:

```julia
    path = joinpath(dir, "REQUIRE")
    isfile(path) && return path
```

to the _below_
https://github.com/JunoLab/Atom.jl/blob/e85b11436edeea25b4c3f8641d013df107ba2734/src/misc.jl#L39-L40

***

Or you meant activating an just really empty environment ? (i.e. without REQUIRE or Project.toml)

pfitzseb:
> Or you meant activating an just really empty environment ? (i.e. without REQUIRE or Project.toml)

Yeah, something like 
```
julia> isfile("Project.toml")
false

(v1.2) pkg> activate .
Activating new environment at `~/.julia/dev/Gumbo/test/fixtures/Project.toml`

(fixtures) pkg> add Example
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
  Updating `~/.julia/dev/Gumbo/test/fixtures/Project.toml`
  [7876af07] + Example v0.5.3
  Updating `~/.julia/dev/Gumbo/test/fixtures/Manifest.toml`
  [7876af07] + Example v0.5.3

julia> isfile("Project.toml")
true
```

aviatesk:
ah yeah, this implementation won't work as expected in that case since it will recur into the parent project file anyway (in that case Gumbo/Project.toml).

I can think we can create an new command for this while keeping the previous command as is (or make config to control the behaviour), but there will be "too many commands" again ...

aviatesk:
@pfitzseb how about creating `julia-client: activate-environment-in-parent-folder` ?

pfitzseb:
Yeah, sounds good.

aviatesk:
done by https://github.com/JunoLab/atom-julia-client/pull/670 :)

PR #242, opened 2020-01-14
access Base.ARGS directly

pfitzseb:
part of the fix for https://github.com/JunoLab/Juno.jl/issues/450.

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -5,7 +5,9 @@ handle("ping") do
 end
 
 handle("args") do args
-  append!(ARGS, args)
+  if length((!isempty).(args)) == 0
+    append!(Base.ARGS, args)
+  end
 end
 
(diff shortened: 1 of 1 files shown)


src/misc.jl
```diff
@@ -5,7 +5,9 @@ handle("ping") do
 end
 
 handle("args") do args
-  append!(ARGS, args)
+  if length((!isempty).(args)) == 0
+    append!(Base.ARGS, args)
+  end
```
aviatesk: sorry for commenting after merge, but does this need to be 
```julia
if length(filter!(!isempty, args)) !== 0
  append!(Base.ARGS, args)
end
```

?

pfitzseb: Yeah, or `sum` instead of `length`. Will push a commit.

PR #244, opened 2020-01-24
update profiler to use FlameGraphs.jl

pfitzseb:
Requires
- https://github.com/JunoLab/Juno.jl/pull/463
- https://github.com/JunoLab/atom-ink/pull/247
- https://github.com/JunoLab/atom-julia-client/pull/672

Fixes https://github.com/JunoLab/Juno.jl/issues/452.

Feels good to delete all that code and get additional functionality for free :)

Juno's now showing gc/dispatch info as well:
image


timholy:
Coool...

timholy:
Are you using FG 0.1.1? If so, are you interested in https://github.com/timholy/FlameGraphs.jl/pull/6 working for you too?

pfitzseb:
Yes, very much.
[This check](https://github.com/timholy/FlameGraphs.jl/pull/6/files#diff-f93d323704121e667dc8254b16330b52R135) would need to be against an input argument because Juno's inline evaluation doesn't go through the REPL machinery.

aviatesk:
Looks so fancy ! Will take a review on this later (maybe this weekend) :)

timholy:
Would be happy to generalize it. If there's a concrete pattern that you don't think will change in the future we can just hard-wire it (note since it's just a symbol comparison it doesn't add a dependency), but it would also be fine to pass some kind of pattern-matching argument through.

Are you comfortable submitting a PR to just implement what you need? If so I'll give you commit access to FlameGraphs.

src/profiler/profiler.jl
```diff
@@ -1,73 +1,27 @@
...
+    :location => name,
+    :func => node.data.sf.func,
+    :line => node.data.sf.line,
+    :count => length(node.data.span),
+    :classes => classes,
+    :children => [tojson(c) for c in node]
+  )
+end
+
+profiler(data = Profile.fetch() ;kwargs...) = @msg profile(tojson(FlameGraphs.flamegraph(data ;kwargs...)))
```
aviatesk:
```suggestion
function profiler(data = Profile.fetch() ;kwargs...)
  graph = FlameGraphs.flamegraph(data ;kwargs...)
  graph === nothing && return
  @msg profile(tojson(graph))
end
```

will fix no method err when profile data is empty

aviatesk:
Also fixes https://github.com/JunoLab/Juno.jl/issues/428

aviatesk:
How about also colouring inline highlights ?
I implemented locally and looks good.
image

pfitzseb:
Ok, I've implemented the pruning here for now. 
[This](https://github.com/JunoLab/Atom.jl/pull/244/files#diff-7ce539ea85face2f596ba4da99859664R52) could be upstreamed, but the toplevel pruning [here](https://github.com/JunoLab/Atom.jl/pull/244/files#diff-7ce539ea85face2f596ba4da99859664R33) probably can't really.

timholy:
It looks mostly upstreamable, if you'd prefer to do so. Want me to tackle that? This basically tells me what needs to be done.

Or we can leave it like this for now. Your choice.

src/profiler/profiler.jl
```diff
@@ -1,74 +1,90 @@
...
+          (
+            c.data.sf.func == :task_done_hook &&
+            endswith(string(c.data.sf.file), "task.jl")
+          ) ||
+          (
+            # REPL
+            c.data.sf.func == :eval &&
+            FlameGraphs.isleaf(c) &&
+            endswith(string(c.data.sf.file), "boot.jl")
+          )
```
aviatesk: this mayn't be so important, but why do we need this ?
Doesn't `pruneinternal!` cut off the branch including `Core.eval` inside of `repleval` anyway ?

pfitzseb: No, it doesn't. I *think* this is an artifact of all the `@async` stuff happening in Juno, but not sure.

aviatesk:
ok, I want to merge those PRs for now once I confirm CI passes here.
@pfitzseb do you still have anything to do within this PR ???

cossio:
Nice! Please tag a release with this.

aviatesk:
I will want to add few more fixes and improvements in next coming 1~2 weeks, so please stay tuned.

PR #251, opened 2020-02-05
Fixing path detection for precompile

aminya:
After I submitted my PR, you edited the code I had which results in an error, so I fixed it (https://github.com/JunoLab/Atom.jl/commit/b60d139c2e1b945101b9c3b807898f13ab84df06).  Also switched to Julia 1.4.0-rc1 for running the script instead of nightly which is not stable enough yet.

- Fixing path detection
- Using Julia 1.4.0-rc1

### DIFF
--- a/.github/workflows/SnoopCompile.yml
+++ b/.github/workflows/SnoopCompile.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Install dependencies
         run: julia --project -e 'using Pkg; Pkg.instantiate();'
       - name : Add SnoopCompile and current package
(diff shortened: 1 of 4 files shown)


aminya:
Here is a successful script run after committing this: 
https://github.com/aminya/Atom.jl/runs/426577065?check_suite_focus=true

This is the PR generated by this:
https://github.com/aminya/Atom.jl/pull/10

aviatesk:
Thanks for crafting this !
I would like to close this in favour of #252 .

PR #252, opened 2020-02-06
update SnoopCompile

aviatesk:


aminya:
Please merge https://github.com/JunoLab/Atom.jl/pull/251. I fixed the tests and depcrecated macros

.github/workflows/SnoopCompile.yml
```diff
@@ -3,13 +3,12 @@ name: SnoopCompile
...
   - push
 
-
 jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['nightly']
+        julia-version: ['1.3']
```
aminya: You should use `1.4.0-rc1` to get the latest SnoopCompile feature.

aviatesk: Is the cutting edge feature important ?
If not, I would prefer generating against the stable release.

aminya: Here it is different. A `1.4` feature is used to remove invalid precompilation sentences. So using older than `1.4` is less stable in this case:
https://github.com/timholy/SnoopCompile.jl/blob/8d73d1f1d70ca2576e3ef3f6f1d0b80080a4973e/src/parcel_snoopi.jl#L276

aviatesk: I see. Thanks !

PR #254, opened 2020-02-06
update outline.jl

aviatesk:
- introduce `OutlineItem` struct
- remove array creations


aviatesk:
Going to refactor whole static features after #262 gets merged.

PR #259, opened 2020-02-06
accept formatting options

aviatesk:
@domluna could you tell me what version-compat we should have in order to guarantee for those formatting options to work ?:
https://github.com/JunoLab/Atom.jl/blob/6d8d35afeb1ebb9bb7ff0083d1b7fac18af9f6aa/src/formatter.jl#L18-L21

Currently we have specification below:
https://github.com/JunoLab/Atom.jl/blob/master/Project.toml#L45


domluna:
@aviatesk should be v0.3.0+

aviatesk:
@domluna thanks ;)

Issue #223, opened 2019-11-24
Function arguments are missed by `locals` when the function body type is declared


aviatesk:
MRE:
```jl
function decltest(ary::Vector{T})::Array{T, 1} where {T<:Number}
    # something
end
function decltest2(ary::Vector{T}) where {T<:Number}
    # something
end
```

and `locals` misses `ary` in `decltest` while it doesn't for that in `decltest2`.

Actually `localbindings` can find `ary` in `decltest`, but [this scope span check in `_actual_bindings`](https://github.com/JunoLab/Atom.jl/blob/avi/renamerefactor/src/static/local.jl#L124) filters out that.

PR #262, opened 2020-02-08
update to latest CSTParser

### LINKED ISSUE #226: update to CSTParser@1.1.0

aviatesk:
Fixes #226 .

Previous tests are almost never modified and so we now can update CSTParser while keeping all the functionalities we've implemented.
The basic binding test is added.

Later I'll add a small fix related to #223 .

TODO:
- [x] test `mark_bindings!` and stuff ?

### DIFF
--- a/src/static/scope.jl
+++ b/src/static/scope.jl
@@ -11,18 +11,21 @@ NOTE:
 function hasscope(x::EXPR)
     t = typof(x)
 
-    # added conditions below when adapted
-    if t === CSTParser.Call && (p = parentof(x)) !== nothing && !hasscope(p)
+    # NOTE: added conditions below when adapted
+    if t === CSTParser.TupleH && (p = parentof(x)) !== nothing && !hasscope(p)
         return true
-    elseif t === CSTParser.TupleH && (p = parentof(x)) !== nothing && !hasscope(p)
-        return true
+    # # XXX:
+    # # introduced in https://github.com/JunoLab/Atom.jl/commit/7de5299001395b83bab0cb9d102e3f36d5c202d1
+    # # but now this seems to include bad cases for function arguments
+    # elseif t === CSTParser.Call && (p = parentof(x)) !== nothing && !hasscope(p)
+    #     return true
     elseif iswhereclause(x)
         return true
     elseif t === CSTParser.MacroCall
         return true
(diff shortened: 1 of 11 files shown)


src/static/scope.jl
```diff
@@ -0,0 +1,107 @@
...
+    t = typof(x)
+
+    # NOTE: added conditions below when adapted
+    if t === CSTParser.TupleH && (p = parentof(x)) !== nothing && !hasscope(p)
+        return true
+    # # XXX:
+    # # introduced in https://github.com/JunoLab/Atom.jl/commit/7de5299001395b83bab0cb9d102e3f36d5c202d1
+    # # but now this seems to include bad cases for function arguments
+    # elseif t === CSTParser.Call && (p = parentof(x)) !== nothing && !hasscope(p)
+    #     return true
```
aviatesk: @pfitzseb 
This scope condition check is introduced in https://github.com/JunoLab/Atom.jl/commit/7de5299001395b83bab0cb9d102e3f36d5c202d1, but now it causes rare case problems like #223 .
This is because `CSTParser.Call` also represents function arguments and if a `LocalScope` is introduced for it, `actual_localbindings` filters it out because of [`span` check](https://github.com/JunoLab/Atom.jl/blob/5962641e3fc61a32d69155676dee0dd6585589ae/src/static/local.jl#L131-L133), in some cases.
Ideally we want to make a span of function argument `Call` equivalent to the span of its `FunctionDef`, but I'm not sure it's really possible, since I think we can't tell the difference between function argument `Call` and usual `Call`s in function body.

If we don't introduce `LocalScope` for it, then function arguments are not filtered out.

***

So could you tell me why you wanted to introduce scopes for `Call`s ? I guess we actually don't need this and tests are passed if we don't have this check.

pfitzseb: This might've been to keep kwargs in function calls out of the local completions/outline. As you said, this doesn't seem to be necessary anymore, so we can in fact remove this check.

pfitzseb:
Great work on this!

PR #264, opened 2020-02-09
inline evlauation during debugging

aviatesk:
addresses:
- https://github.com/JunoLab/Juno.jl/issues/362#issuecomment-569125114
- the second idea of https://github.com/JunoLab/Juno.jl/issues/362#issue-493724338

image:
image

### DIFF
--- a/src/debugger/debugger.jl
+++ b/src/debugger/debugger.jl
@@ -14,6 +14,7 @@ using ..Atom, MacroTools, Lazy, Hiccup
 
 include("breakpoints.jl")
 include("stepper.jl")
+include("eval.jl")
 include("repl.jl")
 include("workspace.jl")
 include("datatip.jl")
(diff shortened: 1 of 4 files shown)


pfitzseb:
What happens if you change and re-eval a function (or even the currently debugged function)? Does the debugger pick up on that?

aviatesk:
> What happens if you change and re-eval a function (or even the currently debugged function)? Does the debugger pick up on that?

If you evaluate the other function (i.e. which is not currently debugged), it won't be evaluated by interpreter by [this check](https://github.com/JunoLab/Atom.jl/blob/avi/debugeval/src/eval.jl#L78).
If you evaluate the function itself, it will be evaluated in interpreter and result in defining a local function within a debugging scope.
(E.g. if we evaluate `strlimit` in the above example, it will introduce `Atom.var"#strlimit#322"()` in the debug scope, and the original (debugged) `strlimit` won't change.)

Well, the check to decide whether evaluate a code in interpreter or our usual eval scheme is in [`interpret_string`](https://github.com/JunoLab/Atom.jl/blob/avi/debugeval/src/debugger/eval.jl#L9-L29), and it has an obvious problem as before -- it can't handle the changes that are not yet evaluated before and after starting debug. But this applies to most of the other debugger features we have.

PR #267, opened 2020-02-09
compat helper

aviatesk:
@pfitzseb how about using compat helper since we have lots of dependencies ?


pfitzseb:
Sure, I guess.

aviatesk:
Let's try how we feel :)

PR #268, opened 2020-02-09
refactor static features

aviatesk:
general:
- more preferable naming

goto:
- recover goto for unloaded packages (can be invoked via docpane)

outline:
- introduce `ActualLocalBinding` struct
- minimum condition check for `toplevelitems`
- show `__precompile__` calls in outline

Needs https://github.com/JunoLab/atom-julia-client/pull/681


aviatesk:
Will merge this once I add a simple test for unloaded package goto.

PR #269, opened 2020-02-10
debug whole file

pfitzseb:
Outstanding issues:
- [x] REPL mode switching is broken.
- [x] We should get rid of unnecessary return statements in top-level code.

Related issues:
- fixes https://github.com/JunoLab/Juno.jl/issues/451
- fixes https://github.com/JunoLab/Juno.jl/issues/425
- fixes https://github.com/JunoLab/Juno.jl/issues/315

Also finally fixes https://github.com/JunoLab/Juno.jl/issues/118.


aviatesk:
Will give a review later.

src/debugger/stepper.jl
```diff
@@ -76,8 +77,79 @@ function add_breakpoint(bp)
...
+            insert!(expr.args, i, JuliaInterpreter.BREAKPOINT_EXPR)
+            insert!(expr.args, i, lnn)
+            i -= 1
+          end
+        end
+      end
+    end
+    if arg isa Expr && !(arg.head in (:function, :struct))
+      insert_bps!(arg)
+    end
```
aviatesk: Guess we need to handle assignment form function definition as well ? Maybe by `MacroTools.isdef(arg)`.

pfitzseb: Yeah, I forgot that short form function definitions don't lower the same...

aviatesk:
Why is there a problem in REPL mode switching ? I still can't see why only toplevel debugging has he issue ...

pfitzseb:
Because it's not started from the REPL, and asynchronously switching REPL modes doesn't work well.

src/debugger/stepper.jl
pfitzseb: Irrelevant now since we use the upstream implementation.

src/debugger/stepper.jl
```diff
@@ -76,8 +77,55 @@ function add_breakpoint(bp)
...
+
+    lc = should_step ? :stepexpr : :continue
+
+    debugmode(true)
+
+    for (i, (mod, ex)) in enumerate(exprs)
+      lc == :stop && break
+
+      temp_bps = add_breakpoint.(Atom.rpc("getFileBreakpoints"))
```
aviatesk: Do we still need this ? Aren't breakpoints added in `startdebugging` anyway for now ?

aviatesk:
> Because it's not started from the REPL, and asynchronously switching REPL modes doesn't work well.

Isn't possible to "pretend" that it's started from REPL when we call `debug_file` ?

src/debugger/stepper.jl
pfitzseb: We do, yeah. Not entirely sure why though :P

aviatesk:
@pfitzseb d4d36ac and https://github.com/JunoLab/atom-julia-client/pull/683/commits/a14acef1fa2d476d83dfa3f76979795c61251813 works nicely as far as I've tested, great work ! Code certainly looks really hacky though.

aviatesk:
I occasionally get the following error:
```julia
RROR: syntax: invisible character \u200b near column 1
Stacktrace:
 [1] top-level scope at none:0
 [2] eval at .\boot.jl:331 [inlined]
 [3] eval_code(::JuliaInterpreter.Frame, ::Expr) at ~\julia\packages\JuliaInterpreter\src\utils.jl:481
 [4] eval_code(::JuliaInterpreter.Frame, ::String) at ~\julia\packages\JuliaInterpreter\src\utils.jl:462
 [5] interpret at ~\julia\packages\Atom\src\debugger\repl.jl:94 [inlined]     
 [6] interpret at ~\julia\packages\Atom\src\debugger\repl.jl:93 [inlined]
 [7] (::Atom.JunoDebugger.var"#65#66")(::REPL.LineEdit.MIState, ::Base.GenericIOBuffer{Array{UInt8,1}}, ::Bool) at ~\julia\packages\Atom\src\debugger\repl.jl:40
 [8] #invokelatest#1 at .\essentials.jl:710 [inlined]
 [9] invokelatest at .\essentials.jl:709 [inlined]
 [10] run_interface(::REPL.Terminals.TextTerminal, ::REPL.LineEdit.ModalInterface, ::REPL.LineEdit.MIState) at ~\julia\julia\usr\share\julia\stdlib\v1.5\REPL\src\LineEdit.jl:2354
 [11] run_interface(::REPL.Terminals.TextTerminal, ::REPL.LineEdit.ModalInterface) at ~\julia\julia\usr\share\julia\stdlib\v1.5\REPL\src\LineEdit.jl:2348
 [12] debugprompt() at ~\julia\packages\Atom\src\debugger\repl.jl:56
 [13] (::Atom.JunoDebugger.var"#53#55")() at .\task.jl:358
```

Do you think you can fix this within the current approach ?

pfitzseb:
That happens when `hideprompt` is called while the debugger (or another repl mode that's not closed with backspace) is active. I thought I'd stopped that from happening with [this line](https://github.com/JunoLab/Atom.jl/pull/269/commits/0614e71969a19ae8b1410544f1ab02a247b7d557#diff-3d1db777e605375c2f73769965f0d125R91), but maybe we should whitelistthe base REPL modes instead.

aviatesk:
Ah I see. Not sure why `isdebugging` filtering doesn't work for me.

aviatesk:
Sorry @pfitzseb, I cluttered with branches and just found I've not included 0614e71 -- will try this again and if the error doesn't happen I guess we can move to release 👍

aviatesk:
Okay, so I found I only see the error when I start debugging within tree-view (which I added in https://github.com/JunoLab/atom-julia-client/pull/684)
Will try to look at why that causes a problem, but I think you can merge for now if you want.

aminya:
Yay. 🚀

PR #274, opened 2020-02-11
prettier current code rendering for frames in unsaved editor

aviatesk:
before:
image

after:
image


aviatesk:
@pfitzseb you want to merge this into yours ?

pfitzseb:
Let's just merge this as-is, there shouldn't be any conflicts.

PR #276, opened 2020-02-13
moar return types:

aviatesk:
- Show return types for all the method completions
  * revert back some part that was introduced in #234 
  * I found revealing a return type on selection is not natural since all the `leftLabel` of suggestions are visible
  * first time method completion invocation is reverted back to be slow, though.
- Now return types can be inferred from input types, while we only used method signature in the previous implementation. Now we can infer return type for more completion cases.
  * E.g.: If we have `f(a) = a`, within the previous behaviour we can't infer in any completion case. With this PR we can show return types if there is input arguments:
    + `f(1|` => inferred as `Int`
    + `f('1'|` => inferred as `Char`

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -121,6 +121,36 @@ completionreturntype(c::REPLCompletions.DictCompletion) =
   shortstr(valtype(c.dict))
 completionreturntype(::REPLCompletions.PathCompletion) = "Path"
 
+using JuliaInterpreter: sparam_syms
+
+function completionreturntype(c::REPLCompletions.MethodCompletion)
+  world = typemax(UInt) # world age
+
+  # first infer return type using input types
+  # NOTE:
+  # since input types are all concrete, the inference result is the best what we can get,
+  # so here we eagerly respect that if inference succeeded
+  f = c.func
+  tt = Base.tuple_type_tail(c.input_types)
+  inf = try
+    Core.Compiler.return_type(f, tt, world)
+  catch err
+    nothing
+  end
+  inf ∉ (nothing, Any, Union{}) && return shortstr(inf)
+
+  # sometimes method signature can tell the return type by itself
+  m = c.method
+  sparams = Core.svec(sparam_syms(m)...)
+  inf = try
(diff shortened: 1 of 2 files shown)


pfitzseb:
> I found revealing a return type on selection is not natural since all the leftLabel of suggestions are visible
> first time method completion invocation is reverted back to be slow, though.

Honestly I think it's better to keep the return types lazy. Looks good otherwise.

aviatesk:
> Honestly I think it's better to keep the return types lazy. Looks good otherwise.

Ah really. Actually to infer using input types lazily is a bit more difficult since we need to "deserialize" input types in create original suggestions ⇒ add details process.
But, okay I will try to do that.

pfitzseb:
Right, I remember that code being fairly complicated. Let's just go with this as-is then, I think.

aviatesk:
well, I just have done the most work in fd0347b.
It works as far as used, certainly it's fairly complicated though.

src/completions.jl
```diff
@@ -232,29 +241,36 @@ function completiondetail_keyword!(comp)
...
-  # HACK: reconstruct a Method object from its hash string
+  fstr, input_types_str, fhash = split(comp[:detailtype], ' ')
+
+  # reconstruct function itself handling gensyms
+  if (m = match(GENSYM_REGEX, fstr)) !== nothing
+    # gensym pattern
+    fsym = Symbol(m.match)
+    ftyp = getfield′(mod, fsym)
+    (!isundefined(ftyp) && isdefined(ftyp, :instance)) || return
+    f = ftyp.instance
```
aviatesk: Hm, so gensym handling here is no so robust, e.g. when a module prefix is also gensymed or so.
I feel it might be better to implement lazy return type inference as a future work, to make a release ?

pfitzseb:
Alright, your call on whether we want the lazyiness or not ;) Feel free to merge with or without that last commit.

aviatesk:
Okay, I would like to merge this as is and include into the next release.
Lazy input type inference may be merged in the future by a separate PR.

PR #278, opened 2020-02-15
RFC: lazy return type inference

aviatesk:
~~Code looks a bit ugly, but works for _most cases_.~~
~~So far I found this can't work for the cases like below:~~
~~- Inference for type constructor method, e.g.: `Dict(:a => 1,`~~

I changed to create a global cache that holds information of `MethodCompletion`, which will be lazily used via `getSuggestionDetailsOnSelect` API and then replaced with the computed details.
Now the lazy completions are achieved while all the `MethodCompletion` power is kept, 
and even more, by reusing computed details cache, we can show return types all at once after we've computed the details
(the cache will get updated when `Method`'s primary world has been changed).


aviatesk:
@pfitzseb I would like to include this into 0.12 if you like this idea. It works fairly good on my machine.

src/completions.jl
```diff
@@ -76,6 +80,8 @@ function basecompletionadapter(
     CompletionSuggetion[]
   end
 
+  length(METHOD_COMPLETION_CACHE) ≥ MAX_METHOD_COMPLETION_CACHE && empty!(METHOD_COMPLETION_CACHE)
```
pfitzseb: Could we make this a FIFO cache instead?

src/completions.jl
```diff
@@ -101,6 +107,87 @@ completion(mod, c, prefix) = CompletionSuggetion(
...
+If the second element is still a `MethodCompletionInfo` instance, that means its details
+  aren't lazily computed yet.
+If the second element is already a `MethodCompletionDetail` then we can reuse this cache.
+
+!!! note
+    If a method gets updated (i.e. redefined etc), key hash for the `REPLCompletions.MethodCompletion`
+      will be different from the previous one, so it's okay if we just rely on the hash key.
+
+!!! warning
+    Within the current implementation, we can't reflect changes that happens in backedges.
```
aviatesk: Maybe the possible problem of this approach is this:
E.g. when we defined
```julia
f() = nothing
g() = f()
```
and invoke method completion for `g` (and cache the inference result `Nothing`),
and after that change `f` like
```julia
f() = 1
```
then, the rt still shows `Nothing`.

I don't think we need to be so strict on this (as in Julia's internal) though.

src/completions.jl
aviatesk: FIFO needs the cache to be `Vector` instead of `Dict`, doesn't it ?
Or we can hold additional information that represents when a items is `push`ed in the `Dict` and implement FIFO.

pfitzseb: We can make it an `OrderedDict` (from OrderedCollections.jl), no?

src/completions.jl
pfitzseb: I think that's fine for now. I feel like it should be possible to fix this, but we can always do that later.

src/completions.jl
aviatesk: Ah yeah, definitely. I didn't know we are using that haha

PR #283, opened 2020-02-17
try catch in _precompile_

pfitzseb:
Not sure why the workspace tests fail _only_ when run while generating the precompile statements...

### DIFF
--- a/scripts/generate_precompile.jl
+++ b/scripts/generate_precompile.jl
@@ -5,18 +5,18 @@ using Pkg
 
 # update when an assertion fails
 blacklist = [
-    "precompile(Tuple{typeof(Atom.processval!),Any,String,Array{Any,1}})"
-    "precompile(Tuple{typeof(Atom.processval!),Function,String,Array{Any,1}})"
-    "precompile(Tuple{typeof(Atom.render′),Juno.Inline,Type})"
-    "precompile(Tuple{typeof(Atom.wsicon),Module,Symbol,Any})"
-    "precompile(Tuple{typeof(Atom.wstype),Module,Symbol,Any})"
-    "precompile(Tuple{typeof(Base.Broadcast.broadcasted),Function,Array{Atom.GotoItem,1},Function})"
-    "precompile(Tuple{typeof(Base.allocatedinline),Type{Atom.GotoItem}})"
-    "precompile(Tuple{typeof(Media.render),Juno.Inline,Type})"
-    "precompile(Tuple{typeof(getfield′),Any,String,Atom.Undefined})"
-    "precompile(Tuple{typeof(getfield′),Any,String})"
-    "precompile(Tuple{typeof(getfield′),Any,Symbol,Atom.Undefined})"
-    "precompile(Tuple{typeof(getfield′),Any,Symbol})"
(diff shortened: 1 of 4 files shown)


scripts/generate_precompile.jl
```diff
@@ -34,6 +34,7 @@ toml = Pkg.TOML.parsefile(project_file)
 test_deps = get(toml, "extras", nothing)
 
 @info "Adding temporary dependencies ..."
+Pkg.activate(@__DIR__)
```
aviatesk: I think this doesn't work, then we can't use dependencies of Atom.jl like `JSON` etc in `runtests.jl`.
This is why I choose to temporarily modify `Project.toml` of Atom.jl each time.

aviatesk:
> Not sure why the workspace tests fail _only_ when run while generating the precompile statements...

Maybe because the version of `CodeTools` of your _Atom.jl's Manifest_ isn't the latest ?
Works on my env.

aviatesk:
I just found those `try/catch` blocks somehow removes all the precompille errors -- even I removed all the blacklists there were no `@debug` messages. Not sure, but does this mean they actually don't work ?

pfitzseb:
I'm getting 
```
λ JULIA_DEBUG=Atom j1 -q
julia> using Atom
[ Info: Precompiling Atom [c52e3926-4ff0-5f6e-af25-54175e0327b1]
┌ Debug: AssertionError("precompile(Tuple{typeof(Atom.processval!), Any, String, Array{Any, 1}})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:338
┌ Debug: AssertionError("precompile(Tuple{typeof(Atom.wsicon), Module, Symbol, Any})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:518
┌ Debug: AssertionError("precompile(Tuple{typeof(Atom.wstype), Module, Symbol, Any})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:632
┌ Debug: AssertionError("precompile(Tuple{typeof(Base.Broadcast.broadcasted), Function, Array{Atom.GotoItem, 1}, Function})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:674
┌ Debug: AssertionError("precompile(Tuple{typeof(Media.render), Juno.Inline, Type})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:1010
┌ Debug: AssertionError("precompile(Tuple{typeof(getfield′), Any, String, Atom.Undefined})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:1064
┌ Debug: AssertionError("precompile(Tuple{typeof(getfield′), Any, String})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:1070
┌ Debug: AssertionError("precompile(Tuple{typeof(getfield′), Any, Symbol, Atom.Undefined})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:1076
┌ Debug: AssertionError("precompile(Tuple{typeof(getfield′), Any, Symbol})")
└ @ Atom ~/.julia/dev/Atom/src/precompile.jl:1082
```
with the latest commit, so everything seems to work fine.

aviatesk:
Ah I forgot all the `@assert`ion things. Let's merge this then.

PR #285, opened 2020-02-17
test package on julia 1.1

jebej:
Tests are failing on 1.0, for now add testing on 1.1 only

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,14 @@ os:
   - osx
 
 julia:
-  - 1.1
+  - 1.0
   - 1.3
   - 1.4
   - nightly
 
 matrix:
   allow_failures:
+    - julia: 1.0
     - julia: 1.4
     - julia: nightly
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
   - osx
 
 julia:
-  - 1
+  - 1.1
   - 1.3
   - 1.4
   - nightly


.travis.yml
```diff
@@ -6,6 +6,7 @@ os:
 
 julia:
   - 1
+  - 1.0
```
aviatesk: Let's change these lines into a single `- 1.0` since currently `1` is only an alias for `1.3`.

jebej:
Ok there are some deeper problems on 1.0. I'll have to install that version locally to see if I can fix them.

aviatesk:
> Ok there are some deeper problems on 1.0. I'll have to install that version locally to see if I can fix them.

yeah, very much appreciated.
The current error is caused at this line:
https://github.com/JunoLab/Atom.jl/blob/22866b55dfa41848bb1f2242dea0c071d8636397/src/docs.jl#L102

I introduced this branching since `stdlibs` is only defined ≥v1.4, but maybe v1.0 doesn't even have `stdlib`. If you find any alternative for this in `v1.0`, it would solve this problem.

As for return type inference for autocomplete, it may be difficult to find a problem since it very much relies on the internal changes, so I'm okay with to leave it broken.

jebej:
If I manage to fix this I'll make another PR. Maybe the package should be lower bound to 1.1 in the meantime.

aviatesk:
> If I manage to fix this I'll make another PR

Okay, looking forward to the PR !

PR #297, opened 2020-03-04
fix test failures on v1.0

aviatesk:


aviatesk:
In this PR I've not tried to repro segfaults on v1.0 (nor I have v1.0 on my machine, for now), but just tried to fix tests failures on v1.0 and now they are all green.

PR #298, opened 2020-03-09
Fix for custom loggers

pfitzseb:
and improved error handling.

### DIFF
--- a/src/progress.jl
+++ b/src/progress.jl
@@ -59,20 +59,23 @@ function Logging.handle_message(j::JunoProgressLogger, level, message, _module,
     end
   else
     previous_logger = Logging.global_logger()
-    if (Logging.min_enabled_level(previous_logger) <= Logging.LogLevel(level) ||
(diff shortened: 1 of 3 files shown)


src/progress.jl
```diff
@@ -59,20 +59,23 @@ function Logging.handle_message(j::JunoProgressLogger, level, message, _module,
...
 
 Logging.shouldlog(::JunoProgressLogger, level, _module, group, id) = true
 
 Logging.catch_exceptions(::JunoProgressLogger) = true
 
-Logging.min_enabled_level(j::JunoProgressLogger) =
-  min(Logging.min_enabled_level(Logging.global_logger()), Logging.LogLevel(-1))
+function Logging.min_enabled_level(j::JunoProgressLogger)
+  min(Base.invokelatest(Logging.min_enabled_level, Logging.global_logger()), Logging.LogLevel(-1))
+end
```
c42f: I think I can see why you want to do this, to allow the global logger to be switched out at any time.

But yikes; it's really not the way I imagined this would work!

It's an odd situation which doesn't normally occur; it's like the only valid place to call `global_logger` is at top level.

PR #301, opened 2020-03-25
add a timer fallback to hideprompt

pfitzseb:


src/repl.jl
```diff
@@ -133,6 +136,18 @@ function hideprompt(f)
   if INIT_COMPLETE[]
     # Escape REPL modes, then write sentinel and trigger chars.
     can_write_to_terminal = something(Atom.@rpc(writeToTerminal("\b\a$(REPL_SENTINEL_CHAR)$(REPL_TRIGGER_CHAR)")), true)
```
aviatesk: Do we still need this line ? Can't we always `blockinput` for this pass ?

aviatesk:
Honestly speaking, now repl.jl is kinda beyond my understanding, so I don't feel I can give a good code review on this.
Well, I tested this branch for a while today and haven't encounter any issue.

src/repl.jl
pfitzseb: Ah, good point -- we should only start the timer if `can_write_to_terminal == true`.

PR #305, opened 2020-03-27
fix functor display

pfitzseb:
Fixes https://github.com/JunoLab/Juno.jl/issues/535.

### DIFF
--- a/src/display/base.jl
+++ b/src/display/base.jl
@@ -101,10 +101,14 @@ import Base.Docs: doc
 isanon(f) = startswith(string(typeof(f).name.mt.name), "#")
 
 @render Inline f::Function begin
-  name = string(typeof(f).name)
-  inner = match(r"typeof\((.+)\)", name)
-  if inner ≠ nothing
-    name = inner[1]
+  if ismacro(f)
+    name = string(f)
+  else
+    name = string(typeof(f).name)
+    inner = match(r"typeof\((.+)\)", name)
+    if inner ≠ nothing
+      name = inner[1]
+    end
   end
   binding = Docs.Binding(typeof(f).name.module, Symbol(name))
   if isanon(f)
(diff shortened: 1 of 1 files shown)


src/display/base.jl
```diff
@@ -101,9 +101,15 @@ import Base.Docs: doc
...
+  inner = match(r"typeof\((.+)\)", name)
+  if inner ≠ nothing
+    name = inner[1]
+  end
+  binding = Docs.Binding(typeof(f).name.module, Symbol(name))
   isanon(f) ? span(".syntax--support.syntax--function", "λ") :
-    LazyTree(span(".syntax--support.syntax--function", string(typeof(f).name.mt.name)),
-             ()->[(Atom.CodeTools.hasdoc(f) ? [md_hlines(doc(f))] : [])..., methods(f)])
+    LazyTree(span(".syntax--support.syntax--function", name),
+             ()->[(CodeTools.hasdoc(binding) ? [md_hlines(doc(binding))] : [])..., methods(f)])
```
aviatesk: so we need to special case macros since `string(typeof(f).name)` would look weird in those cases:

This snippet would address them:
```julia
@render Inline f::Function begin
  name = string(typeof(f).name)
  inner = match(r"typeof\((.+)\)", name)
  if inner ≠ nothing
    name = inner[1]
  end
  (ism = ismacro(f)) && (name = string(f))
  binding = Docs.Binding(typeof(f).name.module, Symbol(name))
  isanon(f) ?
    span(".syntax--support.syntax--function", "λ") :
    LazyTree(
      span(ism ? ".syntax--string.syntax--quote" : ".syntax--support.syntax--function", name),
      ()->[(CodeTools.hasdoc(binding) ? [md_hlines(doc(binding))] : [])..., methods(f)]
    )
end
```

pfitzseb:
Yeah, I like the other changes you have in #303 (didn't see that until after I pushed this PR though). Feel free to merge this change here into your PR.

PR #308, opened 2020-04-01
fuzzy completions

aviatesk:
should work once https://github.com/JuliaRegistries/General/pull/11930 gets merged.
After that I will refactor the test cases a bit and update precompile stmts.


aviatesk:
examples:
- image
- image

aviatesk:
Alright, I updated precompile statements. It would be nice if I could have a feedback on this :)

src/completions.jl
```diff
@@ -1,6 +1,25 @@
-### baseline completions ###
+import REPL.REPLCompletions, FuzzyCompletions
 
-handle("completions") do data
```
pfitzseb, line 3: Imho we should keep the current handler name for compatibility with older julia-client versions. Maybe have `is_fuzzy` or whatever in `data` instead of two different handlers?

src/completions.jl
```diff
@@ -99,6 +94,100 @@ function basecompletionadapter(
...
+    # general
+    line, mod,
+    # local context
+    context, row - startRow, column,
+    # configurations
+    force
+  )
+end
+
+function fuzzycompletionadapter(
```
pfitzseb, line 125: This seems fairly similar to the `replcompletionadapter` -- maybe we could have an `is_fuzzy` kwarg and do something like
```
repl_provider = is_fuzzy ? FuzzyCompletions : REPLCompletions
repl_provider.completions(...)
```
plus have an `if` block for the different filter methods below?

src/completions.jl
aviatesk: I just prioritized my preference for namings basically 😛. Will change the handler names so that we can preserve the compatibility.

src/completions.jl
aviatesk: hm, then the code after `repl_provider` used would be really type-instable (I confirmed that Julia 1.4 can't infer the type of `repl_provider.completions`, for example)

I haven't taken any actual benchmark, but I would want to endure the verbose code for performance here.

pfitzseb: Hm fair enough then. I don't think the dynamic dispatch would make much of a difference, but let's just keep this as-is then.

aviatesk: > I don't think the dynamic dispatch would make much of a difference

yeah maybe. I may be too nervous about dynamic dispatches..
I will refactor them in the future when I get confidence that dynamic dispatches here don't matter.

src/completions.jl
```diff
@@ -12,40 +31,27 @@ handle("completions") do data
     startRow || 0,
     column || 1,
     # configurations
+    is_fuzzy || true,
```
aviatesk, line 34: Do you think fuzzy completions should be default ?

pfitzseb: Not here. Old julia-client versions probably can't display those correctly, right?

aviatesk: This PR is now compatible with the last version at least (only the fuzziness changes).
Maybe julia-client≦0.11 won't be compatible with this PR but it's same for the current master.

PR #309, opened 2020-04-03
lazy Traceur loading

aviatesk:
Since Traceur is still in alpha (and currently has problems in its performance and robustness), I believe ppl don't use this integration so much that it might be okay that it's not provided by default.
Although Traceur (and Cassette) dependency isn't so heavy, but will improve the loading time for sure.

### DIFF
--- a/src/Atom.jl
+++ b/src/Atom.jl
@@ -4,8 +4,7 @@ __precompile__()
 module Atom
 
 using Base.StackTraces, InteractiveUtils, Logging
-using Juno, Lazy, JSON, MacroTools, Media
-import Requires
+using Juno, Lazy, JSON, MacroTools, Media, Requires
 import Media: @dynamic
 
 function __init__()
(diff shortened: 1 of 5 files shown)


src/Atom.jl
```diff
@@ -38,9 +37,8 @@ function __init__()
     nothing
   end
 
-  Requires.@require WebIO="0f1e0344-ec1d-5b48-a673-e5cf874b6c29" begin
-    include("display/webio.jl")
-  end
+  @require WebIO="0f1e0344-ec1d-5b48-a673-e5cf874b6c29" include("display/webio.jl")
```
aviatesk, line 40: I'm not sure why `webio.jl` is better to be loaded by `@require`. Maybe because building WebIO sometimes fails ?

aviatesk:
@pfitzseb maybe are you against this ?

pfitzseb:
No, I just forgot about this. LGTM.

PR #312, opened 2020-04-07
fix plots for terminal results

pfitzseb:
Fixes https://github.com/JunoLab/Juno.jl/issues/541.

### DIFF
--- a/src/eval.jl
+++ b/src/eval.jl
@@ -36,7 +36,7 @@ function start_eval_backend()
         f, args = take!(eval_channel_in)
         Base.sigatomic_end()
         is_backend_working[] = true
-        res = @errs f(args...)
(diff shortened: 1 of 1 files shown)


aviatesk:
@pfitzseb this certainly fixes the issues, but can I ask you why `invokelatest` here fixes the problem ? Maybe `Plots` etc will later add `show` methods but within the previous implementation they're not called ?

pfitzseb:
It doesn't. That `invokelatest` is just to make sure the `evalshow`/`evalall`/`eval` methods are correctly update when you re-eval them. The only relevant fix here is https://github.com/JunoLab/Atom.jl/pull/312/files#diff-8cda017c802f5654612ec039decb6c97R81.

aviatesk:
ah I see. I missed the addition.

PR #314, opened 2020-04-08
fixes https://github.com/JunoLab/Juno.jl/issues/550

aviatesk:


aviatesk:
It turns out that the thing is just I've never read [this](https://docs.julialang.org/en/v1/manual/style-guide/index.html#Avoid-writing-overly-specific-types-1). Will merge this on if CI passes

PR #315, opened 2020-04-08
internal refactors

aviatesk:
~- introduce `@tojs` macro that enables us to easily construct `NamedTuple`s and `Dict`s~
~- no more verbose `todict` conversion~
~- bit of test robustness improvements~

internal refactors:
- use `NamedTuple` instead of `Dict` where possible: more robust for typo, etc
- no more verbose `todict` conversion
- a bit test improvement

### DIFF
--- a/src/completions.jl
+++ b/src/completions.jl
@@ -1,6 +1,38 @@
 import REPL.REPLCompletions, FuzzyCompletions
 
-const CompletionSuggetion = Dict{Symbol, String}
+const CompletionSuggetion = let
+  n2t = (
+    (:replacementPrefix, String),
+    # suggestion body
+    (:text, String),
+    (:type, String),
+    (:icon, String),
+    (:rightLabel, String),
+    (:leftLabel, String),
+    (:description, String),
+    (:descriptionMoreURL, String),
+    # for `getSuggestionDetailsOnSelect` API
+    (:detailtype, String)
+  )
+  NamedTuple{tuple(first.(n2t)...), Tuple{last.(n2t)...}}
+end
+
+function CompletionSuggetion(
+  prefix, text, type, icon;
+  rl = "", ll = "", desc = "", url = "", detail = ""
+)::CompletionSuggetion
+  return (
+    replacementPrefix = prefix,
+    text = text,
+    type = type,
+    icon = icon,
+    rightLabel = rl,
+    leftLabel = ll,
+    description = desc,
+    descriptionMoreURL = url,
+    detailtype = detail
+  )
+end
 
 # as an heuristic, suppress completions if there are over 500 completions,
 # ref: currently `completions("", 0, Main)` returns 1098 completions as of v1.5
@@ -43,7 +75,7 @@ handle("completions") do data
     context, row - startRow, column,
     # configurations
     force
-  )::Vector{CompletionSuggetion}
+  )
 end
 
 # NOTE:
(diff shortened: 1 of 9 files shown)


src/comm.jl
```diff
@@ -190,3 +190,46 @@ end
...
+
+`ex`:
+- will be transformed into `NamedTuple` with parentheses: `@tojs (k: v)` is `((k = v,))`
+- will be transformed into `Dict` with braces: `@tojs {k: v}` is `Dict(:k => v)`
+- will be transformed into `NamedTuple` without parentheses and brances: `@tojs k: v` is `((k = v,))`
+- supports multiple property specs: `@tojs {k1: v1, k2: v2, ...}` is `Dict(:k1 => v1, :k2 => v2)`
+- supports [JS-like shorthand property notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer): `@tojs sym` is `((sym = sym,))`
+- can accept `=` instead of `:` notation: `@tojs {k = v}` is `@tojs {k: v}`
+"""
+macro tojs(ex)
```
pfitzseb, line 213: Wouldn't it make sense for this to produce a JSON string straight away?

aviatesk: yep, it would be fine. I don't think we want to use this macro when we want to create `Dict` after all.

aviatesk: wait, but if we parse them into JSON string in `@tojs`, then [`json` call in `msg`](https://github.com/JunoLab/Atom.jl/blob/5b85f56366166fb41855979b59a5720067b6faae/src/comm.jl#L129) will produce duplicated strings and it will be really breaking.

aviatesk:
So I guess the naming `@tojs` can be misleading. How about `@obj` ?
Well, we don't necessarily introduce this macro, but I believe this macro can make code cleaner (and a bit more robust because of the beauty of `NamedTuple`) in some cases.

src/comm.jl
pfitzseb: No, we can wrap it in a `JSONText`.

aviatesk: ah I get, thanks for the suggestion

aviatesk: So after all, I found that I don't want the macro to produce JSON string. There are lots of cases where I want to refer to its fields after we constructed those "will be passed to JS" objects (e.g. tests or code body of goto.jl etc).
So in other word, what I really wanted was an easy constructor of `NamedTuple`s ... And so there might be no need of this "JS-like" macro..

aviatesk:
@pfitzseb do you want to have the JS-like `NamedTuple` constructor ? Otherwise I'd like to merge this as is.

src/completions.jl
```diff
@@ -1,6 +1,38 @@
...
+    (:icon, String),
+    (:rightLabel, String),
+    (:leftLabel, String),
+    (:description, String),
+    (:descriptionMoreURL, String),
+    # for `getSuggestionDetailsOnSelect` API
+    (:detailtype, String)
+  )
+  NamedTuple{tuple(first.(n2t)...), Tuple{last.(n2t)...}}
+end
```
pfitzseb, line 18: At this point we should probably just make `CompletionSuggestion` an actual type.

aviatesk: could you tell me why you think so ?

aviatesk: okay, I didn't know JSON.jl can correctly `json` on `struct`s.
I removed those verbose `NamedTuple`s now.

aviatesk:
fwiw, this regex was useful when converting `obj[:property]` accessor to `obj.property` accessor:

```
find> `\[\:(((?!\[).)+)\]
replace> .$1
```

PR #317, opened 2020-04-09
allow more JuliaFormatter options

aviatesk:
once we've done https://github.com/domluna/JuliaFormatter.jl/issues/184, we will no longer need this kind of update

### DIFF
--- a/src/formatter.jl
+++ b/src/formatter.jl
@@ -36,14 +36,18 @@ handle("format") do data
   ))
 end
 
+# HACK: extract keyword arguments of `format_text`; `Base.kwarg_decl` isn't available as of v1.0
 const FORMAT_TEXT_KWARGS = let
   ms = collect(methods(format_text))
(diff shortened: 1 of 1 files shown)


src/formatter.jl
```diff
@@ -8,16 +8,50 @@ handle("format") do data
...
   ))
 end
+
+# HACK: extract keyword arguments of `format_text`; `Base.kwarg_decl` isn't available as of v1.0
+const FORMAT_TEXT_KWARGS = let
+  ms = collect(methods(format_text))
+  filter!(m->m.module==JuliaFormatter, ms)
+  m = match(r";(.+)\)", string(first(ms)))
+  m === nothing ? Symbol[] : Symbol.(strip.(split(m.captures[1]), Ref((' ', ','))))
+end
```
aviatesk, line 45: If you have the other way to do this, let me know ;)

PR #319, opened 2020-04-12
use GitHub Actions:

aviatesk:
- test on all the platforms, machine architecture, with the power of Azure in the back
- fix test on 32bit system
- skip test of `realpath` issue in macOS with v1.0


aviatesk:
So Azure's concurrent task ability allows us to [finish those 15 tests in ~8mins](https://github.com/JunoLab/Atom.jl/actions?query=workflow%3ACI).
Right now I couldn't setup codecov report correctly though. The results are uploaded to codecov server, but uploading the coverage results into PR comment fails with the following error:
```
RROR: HTTP.ExceptionRequest.StatusError(400, "POST", "/upload/v2?&service=custom&commit=b13caf048447000a4411eceea923c158703ff673&slug=JunoLab/Atom.jl&token=", HTTP.Messages.Response:
"""
HTTP/1.1 400 Bad Request
Server: nginx/1.13.12
Date: Sun, 12 Apr 2020 05:44:43 GMT
Content-Type: text/html; charset=UTF-8
Content-Length: 32867
Connection: keep-alive
X-Xss-Protection: 1; mode=block
Content-Security-Policy: default-src 
```

PR #325, opened 2020-04-20
.JuliaFormatter.toml configuration file support

aviatesk:
corresponds to: https://github.com/JunoLab/atom-julia-client/pull/735
needs: https://github.com/JunoLab/atom-julia-client/pull/735


src/formatter.jl
```diff
@@ -1,39 +1,61 @@
...
+      if style == "yas" && isdefined(JuliaFormatter, :YASStyle)
+        config_dict["style"] = JuliaFormatter.YASStyle()
+      end
+    end
+    return kwargs(config_dict)
+  end
+
+  overwrite_options(options, config) = kwargs(merge(options, config))
+end
+
```
aviatesk, line 30: If we're okay to just specify `JuliaFormatter = "^0.4.3"`, we don't need this backward compat check

aviatesk: I don't think there is a motivation to use JuliaFormatter<4.3, so going to give up this.

PR #327, opened 2020-04-23
make svgs pan-/zoomable

pfitzseb:


pfitzseb:
Because I thought we'd need to for Gadfly's interactivity to work. It's very likely I was wrong though, and Gadfly still works interactively with this PR because we're not hitting this code path anyways.

PR #340, opened 2020-07-13
Informative closure rendering

yha:
Also fixes JunoLab/Juno.jl#596 by avoiding `repr` call with anonymous function.

### DIFF
--- a/src/display/base.jl
+++ b/src/display/base.jl
@@ -104,32 +104,24 @@ import Base.Docs: doc
 
 isanon(f) = startswith(string(typeof(f).name.mt.name), "#")
 
-function render_anon(i::Inline, f)
-    r(x) = render(i, x)
-    meths = methods(f)
-    if length(meths) == 1
-        sig, link = view(first(methods(f)))
-        header = span(c(r(sig), " at ", r(link)))
-        meths = ()
-    else
-        header = c("λ", fade(" with $(length(meths)) methods"))
-    end
-    LazyTree( span(".syntax--support.syntax--function", header),
-      ()->vcat([methods_table(i,meths)], _fieldlist(f))
-    )
+# functions where `!hascompactrepr(f)` may display closed-over variables in
+# `repr(f)` which might take a very long time
+function hascompactrepr(f)
+  mt = typeof(f).name.mt
+  # this condition is copied from the implementation of Base.show_function
+  isdefined(mt, :module) && isdefined(mt.module, mt.name) && getfield(mt.module, mt.name) === f
 end
+_displayname(f) = isanon(f) ? "λ" : hascompactrepr(f) ? repr(f) : string(nameof(f))
 
 @render i::Inline f::Function begin
-  if isanon(f)
-    render_anon(i,f)
-  else
-    name = repr(f)
-    name_without_module = string(f)
-    binding = Docs.Binding(typeof(f).name.module, Symbol(name_without_module))
-    LazyTree( span(".syntax--support.syntax--function", name),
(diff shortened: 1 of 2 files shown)


src/display/base.jl
```diff
@@ -102,15 +102,17 @@ import Base.Docs: doc
...
-  name = repr(f)
-  name_without_module = string(f)
-  binding = Docs.Binding(typeof(f).name.module, Symbol(name_without_module))
+@render i::Inline f::Function begin
   if isanon(f)
-    span(".syntax--support.syntax--function", "λ")
+    sig, link = view(first(methods(f)))
+    r(x) = render(i, x)
+    defaultrepr(f, true, span(".syntax--support.syntax--function",
+                              c("function ", r(sig), " at ", r(link))))
```
pfitzseb, line 110: Technically, anonymous functions can have multiple methods, so maybe we should just use the normal rendering code here with a `λ` as the function name?

aviatesk:
@yha good catch, I didn't think `show` method can be so slow for closures, thanks. Well I would like to keep the `λ`-style representation of anonymous functions.

src/display/base.jl
yha: I think the link to the definition line is pretty useful, and having the function signature is nice too. And anonymous function would almost always have exactly one method. Maybe add the methods as a subtree when there is more than one?

yha:
@aviatesk What do you mean by "λ-style representation"? Just having the letter "λ" rather than the word "function"?

aviatesk:
yeah, and don't not show gensyms.

yha:
Yeah, I guess the gensym is not useful. So maybe something like "λ(param1,param2) at filename.jl:20"?

aviatesk:
That looks good to me.

src/display/base.jl
pfitzseb: What do you think about something like
```
@render Inline f::Function begin
  if isanon(f)
    LazyTree(
      span(".syntax--support.syntax--function", "λ"),
      ()->methods(f)
    )
  else
    name = repr(f)
    name_without_module = string(f)
    binding = Docs.Binding(typeof(f).name.module, Symbol(name_without_module))
    LazyTree(
      span(".syntax--support.syntax--function", name),
      ()->[(CodeTools.hasdoc(binding) ? [md_hlines(doc(binding))] : [])..., methods(f)]
    )
  end
end
```

That renders as 
image

yha: That would defeat the purpose of the PR, because it doesn't show the function's fields (variables closed over)

pfitzseb: I didn't actually realize that happened because I only tried non-closue anon functions :)

I think I'd still prefer a method list display as well though.

yha:
Now it looks like this:

image

aviatesk:
Looks good, while I want to keep the mothod list rendering to be consistent with usual functions -- i.e. always show only the `λ` on the head even if we have one single method, and append method lists and closure variables in the tree.

yha:
I realize now this is not the right fix, because the problem is really with closures and has nothing to do with whether the function is anonymous. e.g., in
```
function closure()
  data = ...
  foo(i) = data[i]
end
c = closure()
```
`c` is not anonymous, but `repr` would try to print `data` and hang even after this fix. I'll come up with something else.

yha:
I unified the anonymous and non-anonymous rendering, and am using some ugly hack to check whether `Base.repr` might try to print closed-over variables.
So now `λ` is used as anonymous function name and for "anonymous methods" where the type name starts is a `var"..."`.

image

pfitzseb:
Thanks for the work on this!

PR #343, opened 2020-07-29
fix stepper display

pfitzseb:
Fixes https://github.com/JunoLab/Juno.jl/issues/599 (I think).

### DIFF
--- a/src/debugger/stepper.jl
+++ b/src/debugger/stepper.jl
@@ -363,7 +363,7 @@ end
 
 function stepview(ex)
   out = if @capture(ex, f_(as__))
-    fname = typeof(f).name.mt.name
+    fname = string(f)
     if Base.isoperator(fname)
       Row(interpose(as, span(".syntax--support.syntax--function", string(" ", fname, " ")))...)
     else


aviatesk:
I think this won't help because I confirmed `stepview` actually gets `ex = :((typeof)(false))` as an input for the MRE below:
```julia
let
    a = rand(Bool)
    t = @async begin
        a ? sleep(0) : sleep(1)
        sin(a)
    end
    fetch(t)
end
```

I guess the problems rather lives in [`nextstate`](https://github.com/JunoLab/Atom.jl/blob/master/src/debugger/stepper.jl#L343-L362).

pfitzseb:
Ah, true. The expanded code is
```
   1  2  1 ─       Core.NewvarNode(:(t))
   2* 2  │         a = (rand)(Bool)
   3  362  │  │ %3  = var"#20#22"
   4  362  │  │ %4  = (typeof)(a)
   5  362  │  │ %5  = (Core.apply_type)(%3, %4)
   6  362  │  │       #20 = %new(%5, a)
   7  362  │  │ %7  = #20
   8  362  │  │       task = (Task)(%7)
   9  363  └──│       goto #3 if not false
  10  364  2 ─│       (push!)(Main.:(var"##sync#33"), task)
```
so the `typeof(...)` call does make sense.

Anyhow, this does fix a real bug though with the `Task(...)` printing.

aviatesk:
aha, I see. Let's merge.

PR #355, opened 2020-10-16
fix completion performance regression on v1.6:

aviatesk:
see https://github.com/JuliaLang/julia/issues/38059 for the rationale


aviatesk:
failures on nightly aren't related to this PR, but to our copy-and-paste from Revise

PR #359, opened 2020-11-28
Remove dependency WebSockets

GregPlowman:
Can `WebSockets` be removed as a dependency?
Naively appears not to be used, and functionality seems to be replaced in `HTTP`?

`WebSockets` is preventing `HTTP` from upgrading to `v0.9`


aviatesk:
It's used within WebIO:
https://github.com/JuliaGizmos/WebIO.jl/blob/136fa0baaed1999001f62b4d20e0a46f4f2da8f4/src/WebIO.jl#L121-L123

GregPlowman:
Ha ha.
Oh well, thanks.
Closing now.

PR #371, opened 2021-08-09
update to https://github.com/JuliaLang/julia/pull/38791

aviatesk:


pfitzseb:
Should come up with a better UX for in-editor completions as well.

aviatesk:
yeah, maybe a dedicated command ?

pfitzseb:
Ref https://discourse.julialang.org/t/tab-completion-to-find-methods-for-given-variable-s/66076 :)

PR #388, opened 2026-02-26
`FUNDING.yml`: Remove the old link (Salsa Labs), and point to GitHub Sponsors instead (for the JuliaLang organization)

DilumAluthge:
I know that this repo is not active, but I think it would still be good to fix the outdated funding link.


DilumAluthge:
cc: @ViralBShah @vchuravy

aviatesk (approved):

